Issue #104, opened 2021-10-15
issue: Sets to be removed in v3


Aviortheking:
### What seems to be the issue

Set

### Field(s)

all

### Description

Set to be removed: Reason
- misc/Jumbo: Jumbo is now a variant
- Base/W Promotional: WPromo is now a variant
- Black & White/Radiant Collection: Radiant Collection isRC cards from Legendary Treasures

PR #109, opened 2021-10-18
fix: Remove Radiant Collection

Aviortheking:
BREAKING CHANGE: this set is not meant to exists as it is RC cards from the Legendary Treasures set


Aviortheking:
Closing as it is not necessary to have at the moment #104

PR #115, opened 2021-10-20
feat: Add serie wide energies

### LINKED ISSUE #59: Add serie-wide energy support
# What do you want to change?

- [x] Serie
- [ ] Set
- [ ] Card

# What change do you want to do?
_Describe the change(s) you want to be added/enhanced_

Generally each series add their own new version of an energy card

the goal would be to add support in the serie file of which energy cards were in this serie  
+ each energy would have the id `{serie-id}-{type.toLowerCase()}` (unblocking #33)

```diff
export interface Serie {
	id: string
	name: Languages
+	energies?: Array<Types>
}
```

Aviortheking:
Fix #59


Aviortheking:
Listing in API managed in #155

PR #139, opened 2021-11-04
feature: Complete german meta text

Maxopoly:
Ace Spec, Poke Body, Poke Power and Goldrenrod Game Corner were changed to the text on the german version of the relevant cards. 'Ancient Trait' doesn't have a translation on the card, so I picked the closest thing that makes sense. Some others were also mixed up/wrong and should be fixed now.


Aviortheking:
Thanks for the PR ! @all-contributors add @Maxopoly for his data

PR #269, opened 2022-04-04
fix: Set name in french is a mistype (Hidden Fates)

JuneTwooo:
Correction fr Destinnées => Destinées


Aviortheking:
Thanks for the PR !

PR #417, opened 2023-01-22
feat: Optimize Database for speed

Aviortheking:
If this pass it will also close #364 

so Comparing performance is done against it as adding Git as a dependency is pretty violent

Benchmark procedure:

Running this script using the `time` command on linux

for the old system
```bash
npm i
cd server
npm i
cd ..
npm run compile
cd server
npm run compile
```

for the new system
```bash
npm i
cd server
npm i
cd ..
npm run validate
cd server
npm run compile
```


Aviortheking:
as of this commit the performance diff is a ~500% amelioration in build time

old:
```bash
real    71m2.834s
user    11m45.706s
sys     69m19.655s
```
new:
```bash
real    14m10.302s
user    14m8.942s
sys     0m16.442s
```

Aviortheking:
Depending on the new perfs with #459 and the PR #460 it will be closed

Aviortheking:
Closed in favor of #459 & #460

PR #420, opened 2023-01-26
fix: modified variants in Pokémon Go and Astral Radiance

calli23:


Aviortheking:
@all-contributors add @calli23 data

PR #459, opened 2024-01-02
feat: Move internals to bunJS

Aviortheking:
Test to move to BunJS to have better performances

it should have:
- Better build perfs
- Better server perfs


Aviortheking:
build time validated
from ~3m30s to ~2m50s ~20% better build time
from ~30ms to ~20ms ~30% better runtime response

PR #460, opened 2024-01-03
feat: Add an updated field to the card field

Aviortheking:


Aviortheking:
This PR is currently blocked by https://github.com/oven-sh/bun/issues/5319

PR #494, opened 2024-06-13
Add missing pokemon card dex ids

Reloadaxe:
Added missing pokemon card dex ids with a script (handmade)


Aviortheking:
@allcontributors add @Reloadaxe for data

PR #526, opened 2024-08-24
fix: Reversal Energy type should be "Special"

pitbuster:
Reversal Energy is a special energy. I think the allowed values are "Special" and "Basic" in any case, but that string may come from a translation (Spanish, probably).


Aviortheking:
Thanks for you contribution !

PR #527, opened 2024-08-25
fix: add missing retreat cost and weakness for some cards

pitbuster:


Aviortheking:
Thanks for your contribution :D

PR #558, opened 2024-10-11
feat: add Japanese SVJL & SVJP

Aviortheking:


thomas-bassett:
@Aviortheking is this still required?

PR #590, opened 2024-11-27
fix: Fixed new added "Apex" Cards

usullu:
Added missing Cards for the PTCG Pocket. Also Edited false Entries.


Aviortheking:
Thanks for the pull request !
Images for the missing card will also be added alongside !

Issue #599, opened 2024-11-30
issue: Twilight Masquerade holo rare cards missing reverse variant


dsnidr:
### What seems to be the issue

Card

### Field(s)

variants.reverse

### Description

The rare Twilight Masquerade cards with holo variants are missing reverse holo variants

PR #600, opened 2024-11-30
fix: Twilight Masquerade holo rare cards are missing reverse variants

dsnidr:
This PR adds the reverse variant for holo rare cards in SV06 Twilight Masquerade

closes https://github.com/tcgdex/cards-database/issues/599


Aviortheking:
Thanks for the contribution !

PR #622, opened 2024-12-18
Fix: evolve from data

### LINKED ISSUE #619: enhancement: Add evolveFrom key into the Genetic Apex set
### Data affected

- [X] Card
- [X] Set
- [ ] Serie
- [ ] Other
- [ ] new Data Type

### Please explain in more details what idea you have

The card data for the Genetic Apex set, currently does not contain the ``evolveFrom`` key

Here's an example for Venusaur: https://api.tcgdex.net/v2/en/cards/A1-003

Adding the key would help a lot developers like me, because I then would not need to search through other sets, to be able to find the proper info. Also (I believe) it would reduce traffic to the website.

Reloadaxe:
Fix #619 

Fixes more missing and invalid evolveFrom from other series


thomas-bassett:
@Reloadaxe please fix the merge conflicts for merge, im aware of the age of this PR, will leave it a couple days if no response I will close

Issue #631, opened 2025-01-05
issue: Incorrect Paldea Evolved variants


dsnidr:
### What seems to be the issue

Card

### Field(s)

variants

### Description

Some variants for the S&V Paldea Evolved set are incorrect

PR #632, opened 2025-01-05
fix: Paldea Evolved variants are incorrect

dsnidr:
Fixes Paldea Evolved's variants (mainly the missing rare rev holos, but a couple other corrections too).


dsnidr:
closes #631

PR #642, opened 2025-01-23
Added evolveFrom tag to all Genetic Apex cards to Fix #619

### LINKED ISSUE #619: enhancement: Add evolveFrom key into the Genetic Apex set

FalconChipp:
Used existing data type to add on all Stage 1 and 2 Pokemon with the correct "evolves from" basic card using syntax; 

evolveFrom: {
		en: "[Basic_Name]"
	},

### DIFF
--- a/data/Pokémon TCG Pocket/Genetic Apex/002.ts
+++ b/data/Pokémon TCG Pocket/Genetic Apex/002.ts
@@ -13,6 +13,9 @@ const card: Card = {
 	hp: 90,
 	types: ["Grass"],
 	stage: "Stage1",
+	evolveFrom: {
+		en: "Bulbasaur"
+	}, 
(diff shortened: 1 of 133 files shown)


Aviortheking:
Awesome, thanks for your contribution !

Issue #643, opened 2025-01-24
issue:Missing variants for sv03-072


ecornely:
### What seems to be the issue

Card

### Field(s)

variants

### Description

As stated on the discord server, I have a french version of sv03-072 in 3 variations: normal, holo and reverse. I think that at least two variants are missing.

PR #644, opened 2025-01-24
Added missing variants

ecornely:
It would be my first contribution, I hope it's supposed to be like that but as I reported the issue #643 I tried to commit the patch that go with and add the 3 variations I have.

I can prove they exists at least with this IMG20250124094842


Aviortheking:
Hi, thanks for your contribution !

You did exactly what needed to be done !

PR #661, opened 2025-02-07
fix: Add normal variant for swsh9-008

Octopixell:
Changed the `normal` variant for card `008` in the `swsh9` set (Brilliant Stars) to `true` as this card also has a normal variant which can be obtained from the Build & Battle boxes.

PR made after a small discussion and confirming message on Discord: https://discord.com/channels/857231041261076491/857231041956413451/1337221782389657663

Hoping this will be a useful first contribution 🌟


Aviortheking:
Yep we talked about this, thanks for your PR !

PR #665, opened 2025-02-16
Feat : added tcgp boosters to the database & API

TidusDotExe:
Need a code review and obviously open to feedbacks. This MR does not rely on an issue but on the RFC #641 Need a code review and obviously open to feedbacks. This MR does not rely on an issue but on the RFC #641


Aviortheking:
Hi, thanks a lot for your work !

I like how you implemented the thing,

for the data part, there is already a lot of file and for boosters I think only adding a new field to the set would make it easier to maintain (as boosters are linked to sets)

It also means less touch to the `totally not` awful compiler I made 

You can look at my POC I made on https://github.com/tcgdex/cards-database/compare/master...feat/tmp to see how I imagined both of the RFCs

PR #669, opened 2025-02-26
fix: Add holo variants for sv03-023 and sv03-027

Octopixell:
Added a holo variant for cards 026 and 027 in the sv03 (Obsidian Flames) set.

The holo variants are obtained from the Charizard EX Premium Collection 

Decided to also add the normal and reverse variants explicitly so card data doesn't break if the defaults ever change. Saw there are other cards that also redefine the defaults.


Aviortheking:
Thanks for your PR !

It will be pushed to prod tomorrow with the latest set update

PR #686, opened 2025-03-14
feat: add energies and cards pt

heliojff:
1. Add Energies Card SVE
2. Add Promos SVE
3. Add Base in pt


kvacquier:
@heliojff why did you removed the base energies ? For me the set SVE is a legal set, with 16 cards, 1-8 are the basic SV energies and the 9-16 are the tera-cristal background ones

PR #713, opened 2025-04-27
fix: Add holo variant for sv03-100

Octopixell:
Added holo variant for card 100 (Greavard) in the sv03 (Obsidian Flame) set.

Holo variant can be obtained through Houndstone EX box.

Added known variants in a verbose matter to prevent issues when defaults change

### DIFF
--- a/data/Scarlet & Violet/Obsidian Flames/100.ts
+++ b/data/Scarlet & Violet/Obsidian Flames/100.ts
@@ -48,7 +48,11 @@ const card: Card = {
 	retreat: 3,
 	regulationMark: "G",
 
-	variants: {},
+	variants: {
+		normal: true,
+		holo: true,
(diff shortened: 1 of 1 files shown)


Octopixell:
I see that my editor has formatted the file as well 😔 I see the project is lacking any type of formatting set up through a prettier config or something similar. Perhaps something we want to do? 

Sorry for the unintended change- let me know if you want me to undo the formatting 👍

Octopixell:
Actually- it's done everything the `.editorconfig` states- but just *also* the stuff Prettier does. Would it be an idea to set up a prettier config for the project? That way something like this doesn't happen that often if at all 👍 

Lemme know what you want me to do- if anything.

Aviortheking (approved):

PR #715, opened 2025-04-29
feat(cards): added variants to Journey Together cards

ViermannAn:
feat(cards): added variants to journey together cards

Added variants based on official cardlist pdf


Aviortheking:
Hey, sorry I merged the PR q bit to quickly but it seems it doesn't build as you have a lot of missing `,` through the changes you made 

ex: Scarlet & Violet/Journey Together/190.ts
```
	variants: {
		"firstEdition": false,
		"holo": true,
		"normal": false
		"reverse": false,
		"wPromo": false,
		}
```
a `,` is missing after the `"normal"` variant and it break the build

PR #781, opened 2025-06-24
Updating rarity values for PTCGP

carve084:
Most of these updates are to promo cards. You can derive the rarity of a promo card by looking at the Shinedust cost to get its first flair.


Aviortheking:
hmm, the thing is that rarity is based on what is displayed on the card, so the Promos cards like the physical releases do not "have" a rarity

Aviortheking (changes requested):
hmm, the thing is that rarity is based on what is displayed on the card, so the Promos cards like the physical releases do not "have" a rarity

PR #786, opened 2025-07-01
Feat: variants_detailed

thomas-bassett:
Following from discussion point: https://github.com/orgs/tcgdex/discussions/744 this PR is to introduce the variants_detailed  field for cards. 

Using destined rivals as an initial POC.


Aviortheking:
Hey, thanks for the PR !

one thing, in the database I think it is better to directly replace `variants` with the new variants & only keep `variant_detailed` in the API results, the goal is to ease maintenance burden & only have way to write it inside each file.

but keep the possibility of having the `old` one (it should be easy to detect, if it's a list then new format, if it's an object then old format)

thomas-bassett:
I see what you mean, so in the API convert both fields over so it only returns the single type rather than both, but in the data for non updated they would still have variants or variants_detailed. I can work on that this afternoon.

Aviortheking:
for the v2 the fields while live together in the API with `variants` working just as before & `variants_detailed` being the new & future standard

but in the database only the `variants` field is kept & migrate to being of type `variants | Array<variant_detailed>`

to not break the current database & allow to update to the new variants

in the `v3`, the API will move `variants_detailed` -> `variants` with the old `variants` being removed

thomas-bassett:
@Aviortheking OK so if I'm understanding correct. API will expose both `variants` & `variants_detailed`  but the database will only have `variants` accepting either type

Aviortheking:
@thomas-bassett yes, with the main goal of migrating every cards in the future

thomas-bassett:
I'm with you now. I will address that comment now

server/compiler/utils/cardUtil.ts
```diff
@@ -52,14 +63,26 @@ export async function cardToCardSingle(localId: string, card: Card, lang: Suppor
...
+		variants_detailed: Array.isArray(card.variants) ? card.variants?.map((variant) => {
+			return {
+				type: translate('variantType', variant.type, lang) as any,
+				size: variant.size ? translate('variantSize', variant.size, lang) as any : translate('variantSize', 'standard', lang) as any,
+				stamp: variant.stamp ? variant.stamp.map((stamp) => {
+					return translate('variantStamp', stamp, lang)
+				}) : undefined,
+				foil: variant.foil ? translate('variantFoil', variant.foil, lang) : undefined
+			}
+		}) : undefined,
```
Aviortheking, line 85: It should be possible to have basic detailed variants from the basic variant format

server/compiler/utils/cardUtil.ts
```diff
@@ -1,11 +1,12 @@
...
-import { CardResume, Card as CardSingle } from '../../../meta/definitions/api'
-import { getSet, setToSetSimple } from './setUtil'
+import {exec} from 'child_process'
+import {Card, Set, SupportedLanguages, Types} from '../../../interfaces'
+import {CardResume, Card as CardSingle} from '../../../meta/definitions/api'
+import {getSet, setToSetSimple} from './setUtil'
 import translate from './translationUtil'
-import { DB_PATH, cardIsLegal, fetchRemoteFile, getDataFolder, getLastEdit, resolveText, smartGlob } from './util'
-import { objectMap, objectPick } from '@dzeio/object-util'
+import {DB_PATH, cardIsLegal, fetchRemoteFile, getDataFolder, getLastEdit, resolveText, smartGlob} from './util'
```
Aviortheking, line 7: Can you try to limit the style change of existing code to a minimal if possible ?

Aviortheking (changes requested):
Awesome, this is getting close to good for me

I just have some feedback on the code if you are alright we it :D

server/compiler/utils/cardUtil.ts
thomas-bassett: Sure makes sense, I will make a convertor for it

thomas-bassett: done, have made some assumptions on how the convertor should behave, for firstEdition and wpromo

Aviortheking (dismissed):
Looks good !

I'll merge it after the main update so that I can start fetching datasets

PR #793, opened 2025-07-16
Feat: Add PMCG2

thomas-bassett:
Added cards for Japanese set: Pokemon Jungle


Aviortheking (changes requested):

PR #794, opened 2025-07-17
Feat: Add PMCG3

thomas-bassett:
Added cards for Japanese set: Mystery of the Fossils


Aviortheking (changes requested):

PR #796, opened 2025-07-17
WIP: Feat: Add PMCG5

thomas-bassett:
Added cards for Japanese set: Leaders Stadium (Gym Heros)


Aviortheking (changes requested):

PR #797, opened 2025-07-17
Feat: Add PMCG6

thomas-bassett:
Added cards for Japanese set: Challenge from the darkness (Gym Challenge)


Aviortheking (changes requested):

PR #798, opened 2025-07-17
Feat: Add Neo1

thomas-bassett:
Added cards for Japanese set: Gold, Silver, to a New world


Aviortheking (changes requested):

PR #802, opened 2025-07-17
Feat: Add VS1

thomas-bassett:
Added cards for Japanese set: Pokemon VS


Aviortheking (changes requested):

PR #803, opened 2025-07-17
Feat: Add E1

thomas-bassett:
Added cards for Japanese set: Base Expansion Pack


Aviortheking (changes requested):

PR #804, opened 2025-07-17
Feat: Add E2

thomas-bassett:
Added cards for Japanese set: The Town on No Map


Aviortheking (changes requested):

PR #809, opened 2025-07-17
WIP: Feat: Add ADV1

thomas-bassett:
Added cards for Japanese set: Advanced Expansion Pack


Aviortheking:
Hey @thomas-bassett, I didn't look at the PR nicely enough, but it seems like there is multiple errors inside the datas you submitted

PR #823, opened 2025-07-19
WIP: Feat: Add PCG10

thomas-bassett:
Added cards for Japanese set: World Champions Pack

### DIFF
--- a/data-asia/PCG/PCG10.ts
+++ b/data-asia/PCG/PCG10.ts
@@ -4,6 +4,7 @@ import serie from '../PCG'
 const set: Set = {
 	id: 'PCG10',
 	name: {
+		en: 'World Champions Pack',
 		ja: 'ワールドチャンピオンズパック'
 	},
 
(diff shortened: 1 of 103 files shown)


Aviortheking:
Hey @thomas-bassett, thanks for adding thoses sets, one big change tho, can you only add languages that appeared in ? mostly the internationnals releases should not appear in the datas

thomas-bassett:
Are we allowed to keep the en translation? Or is it only meant to be the Japanese? Dose it not make sense to have the translations?

Aviortheking:
this is something that will happens later with a linking through different sets, but the way it is currently working, it will create new cards in the english section, even if the card doesn't exists

thomas-bassett:
Ah, I see, is that something that is already in place? i.e can I just move the translations now?

thomas-bassett:
@Aviortheking removed the languages from this PR only. Let me know if all good with this then I will go through and change the others. Also worth nothing if you haven't seen, but this is using the new variants structure from my other pr so will keep them in WIP till that is merged.

PR #891, opened 2025-08-26
feat: Add `set_number` to the card endpoint

Aviortheking:


thomas-bassett:
either currently fixed by other prs or fixed in `v3`

PR #896, opened 2025-08-28
Add support for Poké Ball and Master Ball reverse holo variants

FalconChipp:
Introduces support for Poké Ball and Master Ball reverse holo patterns in card variant definitions, interfaces, API types, and translations. Updates card data, type definitions, translation files, and card utility logic to handle these new variant subtypes for Prismatic Evolutions and future sets.

Example card (Exeggcute #001 – Prismatic Evolutions) updated with pokeball_reverse and masterball_reverse variants.

### DIFF
--- a/data/Scarlet & Violet/Prismatic Evolutions/001.ts
+++ b/data/Scarlet & Violet/Prismatic Evolutions/001.ts
@@ -20,6 +20,19 @@ const card: Card = {
 	types: ["Grass"],
 	stage: "Basic",
 
+	// NEW: reverse variants present in Prismatic Evolutions
+	// - normal reverse
+	// - Poké Ball reverse
+	// - Master Ball reverse
+	// (No standard holo on this common, so holo=false)
+	variants: {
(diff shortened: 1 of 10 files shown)


thomas-bassett:
This is going to be handled by variants_detailed that was just released.

FalconChipp:
You’re right that variants_detailed is the intended path forward. The key part of this PR is expanding subtype (previously only 'shadowless' | 'unlimited') to also support 'pokeball' and 'masterball', so these new reverse holo patterns can actually be represented.

The extra booleans in variants were just for convenience/consistency building onto variants_detailed

thomas-bassett:
This current version of variants is going to be deprecated in v3. I would say that trying to maintain the two sets of data is likely more work than it's worth. As in the database, you can only use one of the versions, either old or detailed, so instead of adding these Booleans someone could convert it to the new structure. I'm personally a fan of not expanding the old way, as this is a natural way of getting users to migrate early.

PR #898, opened 2025-08-30
Fix: Set lookup for series with different folder naming

thomas-bassett:
Introduces getSerieIdenti to handle folder name differences between 'data-asia' and 'international' datasets. Updates all getSets calls to use the correct identifier, ensuring proper set retrieval and sorting for series.


thomas-bassett:
this may have been fixed recently. need to confirm

PR #920, opened 2025-09-11
feat: Expand variant_detailed + sv10.5b variants

itsneeku:
I'm planning on adding all variants found on `pkmn.gg`. The data includes
- The "main" variant of a card, that is the first one found in the set. Depicted as a variant of the "standard set"
- All of a card's variants,  including promotional/product exclusive ones.
- A description of the variant (e.g.: "Pikachu Deck stamp Battle Academy exclusive",  "(Cracked Ice) Found in Sword & Shield Rillaboom Theme Deck", etc.)
- The TCGPlayer Store ID of the variant. _Omitted as of now but could be useful?_

## Changes
- Added two fields to `variant_detailed`
  - description, used to denote where a certain variant was found
  - set, used to denote whether the variant is part of the standard set, parallel set, or neither _(could be renamed to something else?)_
- Added variants to all SV10.5b cards

### DIFF
--- a/data/Scarlet & Violet/Black Bolt/001.ts
+++ b/data/Scarlet & Violet/Black Bolt/001.ts
@@ -56,7 +56,36 @@ const card: Card = {
 
 	thirdParty: {
 		cardmarket: 835903
-	}
+	},
+
+	variants: [{
+		type: "normal",
+		size: "standard",
+		description: "Found in Booster Packs",
+		set: "standard"
+	}, {
+		type: "reverse",
+		size: "standard",
+		description: "Found in Booster Packs",
+		set: "parallel"
+	}, {
+		type: "reverse",
+		size: "standard",
+		foil: "pokeball",
+		description: "Found in Booster Packs",
+		set: "parallel"
+	}, {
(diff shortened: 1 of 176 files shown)


itsneeku:
@thomas-bassett 

> Do we need an “other” for set? If this is an unknown value
Yeah probably not, kinda forgot it was optional lol

I don't like the `set` name, but I havent found any official designations to differentiate standard/parallel other than the etb booklets.

To get rid of the `description` and `set`, what are we thinking of differentiating regular prints (found in booster packs) from other prints (confetti holo found in McDonald's Packs, Play! stamps from prize packs, build&battle-specific prints etc.) using its own field?

thomas-bassett:
The others would just be their own variant. For the McDonald's it would just be: `{ type: "holo", foil:"confetti"}` and the others would just have stamps and such instead. Although McDonald's is a bad example as there is part of its own set.

itsneeku:
Right but then there's no way to figure out which variants are the main ones (regular prints, master set) and which are extras (play stamp, grandmaster+ set)

PR #969, opened 2025-11-10
Add japan exclusive stamp promo

amaurybegood:
## Changes

Add two cards:
- Cramorant (S-P 226) Sword & Shield Promos
- Pikachu (S-P 227) Sword & Shield Promos 

## Details

Hello, this is my first PR, tell me if there is something wrong.
Do I need to also edit S-P.ts?


Aviortheking:
Yes, the card localID is defined by the filename, if it is already defined it means cards are present in other languages

PR #996, opened 2025-11-17
feat: Add Trainer card legality for reprints + fixing some cards with surnames

kvacquier:
## Changes

- Add Trainer card legality cache to share legality status between cards with the same name
- Optimize compilation order: process cards from newest to oldest sets (for cache), then sort final output oldest first
- Process Trainer cards sequentially, other cards in parallel for better performance
- Add character name descriptions to Trainer cards (e.g., "Boss's Orders (Giovanni)")

## Details

This PR implements a Trainer card legality cache system that ensures all Trainer cards with the same English name share the same legal status. When a newer Trainer card with the same name is legal, older versions of that Trainer card will also be marked as legal, which aligns with how Trainer cards work in the Pokémon TCG where same-name Trainers share legality.

The compilation process has been optimized to:
- Process cards from newest to oldest sets during compilation (required for the Trainer cache to work correctly)
- Process Trainer cards sequentially to build and check the cache
- Process other cards (Pokemon, Energy) in parallel for better performance
- Sort the final output by release date (oldest first) to maintain the original order

The cache uses English card names as keys to optimize memory usage, and only stores legality information for Trainer cards.


thomas-bassett:
It was more that this requires a strict ordering of compilation which might be risky long term as its possible someone could also change somthing about the compiler which changes the order unaware that this needed it. 

Thus getting incorrect legality. Is there a way this can be garunted without the strict compilation?

kvacquier:
| Thus getting incorrect legality. Is there a way this can be guaranteed without the strict compilation? 
Would be complicated, since the rule is purely based on if there's a card with the same English name currently legal (on trainer card list), so we need to check check that and I did not find a more efficient way than doing this in that order : a hard request make the compilation way slower, generating updated legal .ts would over-complex CI and would be really hard on rotations (where the current way fix that)

kvacquier:
> Looks good, at the moment we are only using the `en` name this wouldnt work for anything in the `data-asia` folder as that wouldnt have `en` in the card data

AFAIK, in asia, each country as it's own format, legality and rules about reprint (at leasts there's rules for specific, Japan, China and Korea that are different...), so we can not do it the same way

thomas-bassett:
> AFAIK, in asia, each country as it's own format, legality and rules about reprint (at leasts there's rules for specific, Japan, China and Korea that are different...), so we can not do it the same way

That's a fair point. can we specifically lock this to the international langs?

PR #1002, opened 2025-11-20
Feat: POC for linking related objects

thomas-bassett:
This pr is a POC for creating links between objects i.e (series,set,cards). currently it allows the manual linking used to link objects between the data and data-asia folders. it also will automatically link translations defined in its self i.e en object will show links for the translations to de,fr,es and so on if it has that data filled out.

used base set as an example


thomas-bassett:
planned for `v3`

PR #1004, opened 2025-11-23
Fix: Update pricing on a interval

thomas-bassett:
This PR fixes pricing not being updated on an interval. Allows customisation of trigger time defaults to 10 minutes, but this doesn't override the cardmarket and tcgplayers 1 hour min, it will just check every 10 minutes if an hour has passed. 

Still keeps the initial run on server start up


thomas-bassett:
Fix is a fix!

PR #1012, opened 2025-11-30
Feat: Improve card lookup by handling leading zeros in IDs

thomas-bassett:
Updated the 'cards' endpoint to attempt fetching cards by both the original and normalized IDs, with leading zeros removed. This enhances compatibility with different ID formats across sets and ensures existing user links remain functional.


thomas-bassett:
fixed in `v3`

PR #1018, opened 2025-12-03
feat: Refactor variant stamp structure to support details

thomas-bassett:
Updated card variant 'stamp' fields to use objects with 'stamp' and 'detail' properties for event-based stamps, replacing previous string formats. Adjusted TypeScript interfaces and translation files to support the new structure, and updated card utility logic to handle both string and object stamps for translation and display.


Aviortheking:
not sure about this, I mean if someone want to support every stamps in their application they would still need to match the base string with their internal icon ?

What does this add ? (I may have missed the convo, if so srry)

thomas-bassett:
We found thst there where stamps that are identical but are off to the left. Or in the case of the battle academy decks the same stamp with a different number. This was to try help with not having "pikachu-tail-01,pikachu-tail-02" the compiler still flatterns them down to a single string it just makes it a little cleaners on the database side of things.

PR #1019, opened 2025-12-04
Feat: Expose thirdparty ids

thomas-bassett:


server/src/libs/providers/tcgplayer/fallback.ts
```diff
@@ -72,9 +72,9 @@ export async function updateTCGPlayerDatas(): Promise<boolean> {
 export async function getTCGPlayerPrice(card: { thirdParty: { tcgplayer?: number } }): Promise<{
 	unit: 'USD',
 	updated: string
-	normal?: Omit<Result, 'productId' | 'subTypeName'>
-	reverse?: Omit<Result, 'productId' | 'subTypeName'>
-	holo?: Omit<Result, 'productId' | 'subTypeName'>
+	normal?: Omit<Result | 'subTypeName'>
```
Querela: Seems like ` | ` has to be `, `, so `Omit<Result, 'subTypeName'>`. According to test failure.
Same for next two lines.

thomas-bassett: whoops dumb moment, had it like that on the official but not fallback

PR #1025, opened 2025-12-09
fix: Remove old server instructions

Querela:
## Changes

The `server/README.md` contains old instructions that are not needed anymore or don't work at all.
- `npm run compile` does not work in the root as the `compile` script is not defined
- `bun` seems to be missing from the dependency list of `server/package.json`. It could be installed globally but maybe defining it in the `server/package.json` would be good, as `devDependencies` or `peerDependencies` or something? The scripts want to use `bun` but it is somewhat assumed that it exists somewhere. I added a NOTE in the meantime that users might need to install it themselves.

### DIFF
--- a/server/README.md
+++ b/server/README.md
@@ -12,19 +12,15 @@ You will need NPM/NodeJS installed
 
 ```
 npm install
-cd ..
-npm install
-cd server
 ```
 
 2. Compile the database
 
 ```
-cd ..
-npm run compile
-cd server
 npm run compile
 ```
+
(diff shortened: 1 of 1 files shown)


server/README.md
```diff
@@ -12,20 +12,16 @@ You will need NPM/NodeJS installed
 
 ```
 npm install
-cd ..
-npm install
```
thomas-bassett: we should use the `bun run ***` instead of npm

thomas-bassett (changes requested):

PR #1049, opened 2025-12-30
fix(cardmarket): update idReference for Scarlet & Violet

Chester-ttv:
## Changes

- Updated `thirdParty.cardmarket` values for the **Scarlet & Violet** set

## Details

- Verified each Cardmarket product manually
- One commit for this set
- This PR is opened as a **draft** to confirm the approach before continuing with other sets

Your feedback is welcome so I can continue with the other sets 🙂


lamplis:
The GraphQL error 'Cannot return null for non-nullable field Set.serie' appears to be unrelated to the cardmarket ID changes in this PR. The PR only modifies  values in card files and doesn't touch any Set or Serie definitions.

This error typically indicates a data integrity issue where a Set object is missing its serie reference. Since the PR changes don't affect Set definitions, this might be:
1. An intermittent test infrastructure issue
2. A pre-existing problem in the base branch

Could you try re-running the failed test workflows to see if the error persists?

lamplis:
## Issue Resolved ✅

I've identified and fixed the root cause of the "Cannot return null for non-nullable field Set.serie" error.

### Root Cause
The error was **not caused by this PR**. It was a pre-existing bug in the `setToSetSimple` function in `server/compiler/utils/setUtil.ts`. This function creates simplified set objects for cards but was missing the `serie` field, which is required as non-nullable in the GraphQL schema.

### Fix
I've created PR #1071 which adds the missing `serie` field to the `SetResume` objects. Once that PR is merged, this PR should pass all tests.

### Summary
- ✅ Your cardmarket ID changes are correct
- ✅ The GraphQL error was unrelated to your changes  
- ✅ Fix has been submitted in PR #1071
- ✅ This PR should be good to merge after #1071 is merged

Thank you for your contribution!

PR #1084, opened 2026-01-26
[Feat][V3] Graphql rework - REQUEST FOR COMMENT

thomas-bassett:
This PR is a full rework of the GraphQL endpoint and is likely to change. This PR is open for others to build locally and test.

With this version of then GraphQL endpoint, it completely removes the Brief/Full object structure and allows full top down and bottom up searching of every value in the API.

This is likely to have some issues and will need further work and discussion, reach out to me on discord at `@Thomas` (snoogle13) with any issues for me to have a look at.

this version of graphql is available under the v3 endpoint
`/v3/graphql`

### Multilang Support
this adds multi Lang support for `cards, sets and series` with the use of the new `locales` object. You can still request translatable fields in the root object but can add this new object and a selection of language filters to get addition information in one language

example query:

```
query {
  cards(filters: { id: "eq:base1-1" }) {
    id
    locales(langs: ["en", "fr", "de"]) {
      lang
      name
      attacks {
        name
        damage
        effect
        cost
      }
      description
    }
  }
}
```

locales accepts all the same fields from the card object except for fields like `id` as they are the same 

> except for when cards cross the International > Asia boundary this still will not work to translate these as there is no mapping currently to get them. Currently cards must share an `id` for them to be found

example return:
```
{
  "data": {
    "cards": [
      {
        "id": "base1-1",
        "locales": [
          {
            "lang": "en",
            "name": "Alakazam",
            "description": null
          },
          {
            "lang": "fr",
            "name": "Alakazam",
            "description": "Son super cerveau peut effectuer des opérations plus rapidement qu'un super ordinateur. Il a un Q.I. de 5000."
          },
          {
            "lang": "de",
            "name": "Simsala",
            "description": null
          }
        ]
      }
    ]
  }
}
```


thomas-bassett:
fixed in `v3`

PR #1108, opened 2026-02-06
feat: Populate CM links for JP cards en masse

dga711:
*work in progress*

This PR aims to populate the cardmarket links for a large set of the JP cards.
Took the existing linking scripts and created new versions which put effort in cleaning names and finding translations.

For this draft I only ran "SV" and "S" sets.

Not sure yet if the scripts will in the end be part of this PR since their datasources are quite custom and I am not sure we want to keep those here.

Most card links that are skipped are because of missing translations, especially for Items and Trainers.
In the future the translation references PR could help this effort.

There will be some % of wrong ids like in the en sets because of variants etc. But better than nothing?


thomas-bassett:
@dga711 this is still marked as WIP please let me know if its ready or not

PR #1113, opened 2026-02-07
[Feat] Allow sets to have multiple card market ids

thomas-bassett:
Card market can have multiple IDs for a single tcgdex set, this PR allows either a single ID or multiple. We may decide that this isn't necessary and can be ignored, if so

### DIFF
--- a/server/compiler/index.ts
+++ b/server/compiler/index.ts
@@ -23,8 +23,8 @@ const DIST_FOLDER = './generated'
 		await fs.rm(DIST_FOLDER, {recursive: true})
 	} catch {}
 
-	console.log('\n2. Loading informations from GIT')
-	await loadLastEdits()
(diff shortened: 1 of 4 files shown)


server/compiler/index.ts
```diff
@@ -23,8 +23,8 @@ const DIST_FOLDER = './generated'
 		await fs.rm(DIST_FOLDER, {recursive: true})
 	} catch {}
 
-	console.log('\n2. Loading informations from GIT')
-	await loadLastEdits()
+	// console.log('\n2. Loading informations from GIT')
+	// await loadLastEdits()
```
dga711: Probably shouldn't be here

thomas-bassett: whoops

Aviortheking (changes requested):
That would be a breaking change, create another field or keep it only in the database only

PR #1116, opened 2026-02-12
Added field types to pokemon cards missing it in international sets

ThugaPT:
## Changes

Field Type was added where missing, with the correct value


thomas-bassett (changes requested):
please address the merge conflicts @ThugaPT

PR #1159, opened 2026-03-12
FIX: Correct the variants for Surging Sparks

jmandev88:
## Changes

Fixes variations for SS

## Details

Updated 252 cards to fix the variants from the old object, to the new array solution


Butz85 (commented):
Looks good!

 Is the API-docs updated with the `foil`-flag to reflect the type of holo?

PR #1167, opened 2026-03-13
Add new ME variant types & Pricing IDs for ASC

FalconChipp:
## Changes

- Added new variant types for Reverse Foils
- Added pricing IDs for ASC 
- Moved thirdParty ids into variants


thomas-bassett:
> Hi, I don't get why we have both reverse and reverse+enenrgy. It makes 4 variants and I only see 3 basic variants in real life and for a few cards we have "cosmos holo" and "AP stamped version"... Maybe I missed something? { type: "reverse", thirdParty: { cardmarket: 869613, tcgplayer: 675814 } },
> 
> and
> 
> ```
> 	{
> 		type: "reverse",
> 		foil: "energy",
> 		thirdParty: {
> 			cardmarket: 870138,
> 			tcgplayer: 676993
> 		}
> 	},
> ```

in this case the default `reverse` is a mistake. most previous sets have had a normal `reverse` along with the special ones

interfaces.d.ts
```diff
@@ -87,6 +87,8 @@ export interface variant_detailed {
 	 */
 	foil?: 'pokeball' | 'greatball' | 'ultraball' | 'masterball' | 'gold' | 'cosmos' | 'galaxy' | 'starlight' | 'energy' | 'cracked-ice'
 	| 'mirror' | 'league' | 'player-reward' | 'professor-program' | 'tinsel'
+	// Added for new Ascended Heroes foil patterns:
```
thomas-bassett: we can remove this comment

interfaces.d.ts
```diff
@@ -431,4 +433,4 @@ export interface Filter {
 		 */
 		cards: Array<string>
 	}
-}
+}
```
thomas-bassett: formating

thomas-bassett (changes requested):

thomas-bassett (approved):

PR #1181, opened 2026-03-20
Add possibility to search a card with name in all languages at the same time

kvacquier:
## Changes

- Added `anyName` query parameter to the `GET /v2/{lang}/cards` endpoint
- Updated OpenAPI spec to document the new parameter

## Details

The existing `name` filter only searches card names in the requested language. This makes it impossible to find a card if you only know its name in another language (e.g. searching `name=Dracaufeu` on `/v2/en/cards` returns nothing, even though Charizard exists).

The new `anyName` parameter searches across all 18 language card lists and returns the matching cards in the requested language:

```
GET /v2/en/cards?anyName=Dracaufeu     → returns Charizard
GET /v2/fr/cards?anyName=Glurak        → returns Dracaufeu
GET /v2/en/cards?anyName=eq:Pikachu    → exact match across all languages
```

It supports the same operator prefixes as other filters (`like:`, `eq:`, `not:`, `neq:`, etc.) and can be combined with any other query parameter (`set`, `rarity`, `hp`, etc.).

---

Note: since this repository is card data only, you may want to open the PR against the server repo instead if that's tracked separately. Let me know.


advra:
I recommend you make your own secondary search layer on top that which tailor to your needs. This is what I have personally done.

PR #1183, opened 2026-03-22
Fix XY EX names to have a -

arthur-reznik:
closes # 

## Changes

Fixed XY -EX cards to have the correct name with -EX as suffix

### DIFF
--- a/data/XY/Ancient Origins/25.ts
+++ b/data/XY/Ancient Origins/25.ts
@@ -3,12 +3,12 @@ import Set from '../Ancient Origins'
 
 const card: Card = {
 	name: {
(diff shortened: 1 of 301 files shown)


thomas-bassett:
they may also be incorrect then, which might require use to change those or these to match. im going to defer merging till the boss has looked at it 

@Aviortheking please can you take a look

arthur-reznik:
Another bump @Aviortheking if you have the time to check on this. I'd rather have it fixed here than on my patched solution for mismatched/custom data

advra:
This is a very good change. Looking forward to this getting merged in

PR #1190, opened 2026-03-23
fix(data): Radiant Collection (Black & White)

lamplis:
Set: **Radiant Collection**
Series folder: `Black & White`
Changed card files: **0**

### Validation
- [ ] `bun run validate`
- [ ] `cd server && bun run --bun validate`
- [ ] `cd server && bun run compile`
- [ ] Manual review: translations, energy types, TS syntax

### Card images
See follow-up comments with embedded TCGdex assets.


thomas-bassett (changes requested):
there is Legendary Treasures and Legendary Treasures (Radiant Collection) this should be kept and name changed

PR #1361, opened 2026-04-08
fix: add User-Agent header to TCGCSV fetch calls

FalconChipp:
## Changes

- Added TCGCSV_USER_AGENT environment variable to control the User-Agent header sent to TCGCSV
- updateTCGPlayerDatas.ts reads from TCGCSV_USER_AGENT and skips pricing update with a warning if not set
- Preload script reads from TCGCSV_USER_AGENT and exits with an error if not set
- Added TCGCSV_USER_AGENT=TCGDex-localDev/1.0.0 to the local docker-compose.yml

## Details

TCGCSV recently started blocking requests that don't identify themselves via a User-Agent header, causing all TCGPlayer pricing data to return null.
updateTCGPlayerDatas.ts and the TCGPlayer preload script were making fetch calls to tcgcsv.com without a User-Agent header. TCGCSV now rejects these with a 403, which caused the update function to fail silently, leaving lastFetch unset and causing getTCGPlayerPrice to always return null.
The User-Agent is now driven by the TCGCSV_USER_AGENT environment variable rather than being hardcoded, keeping the production identifier (TCGDex/1.0.0) out of source and allowing the dev compose to use a separate identifier (TCGDex-localDev/1.0.0). This prevents local dev traffic from being indistinguishable from production traffic on TCGCSV's end.


scripts/preloadTCGPlayer.ts
```diff
@@ -9,15 +9,22 @@ import { Set } from '../interfaces'
  * Example usage : `bun meta/scripts/check-missing.ts "data/*\/*\/*.ts" thirdParty.cardmarket`
  */
 
+const userAgent = process.env.TCGCSV_USER_AGENT
+if (!userAgent) {
+	console.error('TCGCSV_USER_AGENT is not set')
```
Aviortheking: Can you use a default user agent like tcgdex-dev ?

scripts/preloadTCGPlayer.ts
```diff
@@ -50,4 +59,4 @@ try {
 } catch (error) {
 	console.error('Error:', error.message)
 	process.exit(1)
-}
+}
```
Aviortheking: Keep the last blank line

Aviortheking (changes requested):

scripts/preloadTCGPlayer.ts
FalconChipp: Apologies, missed this off the PR. Was left as is incase you wanted to set a private one through env vars on prod. 
There are changes to the contributing docs to have users set their own to not spam tcgcsv with the same user agent and get us limited

Aviortheking: yep that's what I meant, but instead of closing the script use a default one

FalconChipp: The idea was to not include a fallback if it wasn't set. Meaning a Unique one will need to be set by the user. 

Addressed

scripts/preloadTCGPlayer.ts
FalconChipp: Addressed

PR #1367, opened 2026-04-10
fix(data): normalize legacy trainer text formatting

lamplis:
## Summary
- normalize legacy Trainer effect formatting in Diamond & Pearl, Platinum, XY, and Sun & Moon cards
- restore bullet list structure for French text where the effect branches by option or card count
- replace glued `:•` / `.•` formatting and normalize legacy energy text tokens in the affected Primal Clash cards

## Test plan
- [x] Run `bun run validate`
- [x] Run `cd server && bun run --bun validate`
- [x] Run `cd server && bun run compile`
- [x] Run `cd .bruno && bru run --env Developpement`

### DIFF
--- a/data/Diamond & Pearl/Secret Wonders/121.ts
+++ b/data/Diamond & Pearl/Secret Wonders/121.ts
@@ -14,7 +14,7 @@ const card: Card = {
 	set: Set,
 
 	effect: {
-		fr: "Attachez PlusPower à 1 de vos Pokémon. Défaussez cette carte à la fin de votre tour. Lorsque le Pokémon auquel PlusPower est attachée attaque, cette attaque inflige 10 dégâts supplémentaires au Pokémon Actif (avant application de la Faiblesse et de la Résistance).",
+		fr: "Attachez PlusPower à 1 de vos Pokémon. Défaussez cette carte à la fin de votre tour.\n\nLorsque le Pokémon auquel PlusPower est attachée attaque, cette attaque inflige 10 dégâts supplémentaires au Pokémon Actif (avant application de la Faiblesse et de la Résistance).",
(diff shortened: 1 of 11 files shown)


lamplis:
**Windows CI fix: retry for git-log commands**

The `loadLastEdits()` function runs thousands of `git log` calls in parallel. On Windows runners, some of these calls fail randomly because the OS runs out of process handles.

This commit fixes it by:
- **Retrying failed commands** up to 3 times with increasing wait time (200ms, 400ms, 800ms)
- **Lowering parallel commands on Windows** from 10 to 5 to reduce pressure
- **Logging each file that still fails** after retries, plus a total count at the end
- **Skipping empty lines** from the file list
- **Removing old commented-out code** that was no longer used

This does not change any card data. It only makes the build more reliable on Windows.

thomas-bassett:
while the complier change is great, ive personally not seen the random failures. it shouldnt be combined with a data change. take the compiler into its own PR and then il merge the data changes

thomas-bassett (changes requested):
as above

lamplis:
The build failed only on GitHub for me. I'll proceed as required.

thomas-bassett:
the windows build inst required on github right now due to its slow speed so as long as the others pass your good

lamplis:
I split the compiler change out of this branch as requested.

This PR now contains only the legacy Trainer text formatting updates across the 11 affected data files. I re-based it onto the latest `master` and re-ran the checks locally:

- `bun run validate`
- `cd server && bun run --bun validate`
- `cd server && bun run compile`
- Bruno API suite: 67/67 requests passed, 118/118 assertions passed

I’ll handle the Windows `git log` retry/backoff change in a separate PR.

thomas-bassett (approved):

PR #1370, opened 2026-04-11
fix(build): split windows git-log retry from #1367

lamplis:
## Summary
- move the Windows `git log` retry/backoff logic into a standalone compiler-only PR
- keep the change scoped to `server/compiler/utils/util.ts` (`loadLastEdits`)
- reduce `win32` concurrency and keep retry diagnostics for files that still fail

## Test plan
- [ ] `bun run validate`
- [ ] `cd server && bun run --bun validate`
- [ ] `cd .bruno && bru run --env Developpement`


Aviortheking:
What does this fix ?
The files currently failing on windows are the same as the ones on the others OSs

lamplis:
> What does this fix ? The files currently failing on windows are the same as the ones on the others OSs

From time to time, the github build fails and I have to re trigger a build.
I had the problem 2 times in a row then I proposed this fix.

thomas-bassett:
this should be fixed in `v3`

PR #1371, opened 2026-04-11
fix(database): Correct Drampa holo variants for McD 2022 and 2024 promos

### LINKED ISSUE #1357: issue: Incorrect Variants on Drampa Cards (example)
### What seems to be the issue

Card

### Field(s)

Variant

### Description

The variants of some cards have extras. 
-'22 Mcdonalds Promo was only a non holo.
-'24 promo was only a holo

https://github.com/tcgdex/cards-database/blob/master/data/McDonald's%20Collection/McDonald's%20Collection%202022/14.ts

Image
Image

santichausis:
McDonald's 2022 Drampa was only released as non-holo (normal: true, holo: false).
McDonald's 2024 Drampa was only released as holo (normal: false, holo: true).

Fixes #1357

## Changes

- **McDonald's Collection 2022 #14 (Drampa)**: Set `holo: false` — this promo was only released as a non-holographic card.
- **McDonald's Collection 2024 #15 (Drampa)**: Set `normal: false` — this promo was only released as a holographic card.

### DIFF
--- a/data/McDonald's Collection/McDonald's Collection 2022/14.ts
+++ b/data/McDonald's Collection/McDonald's Collection 2022/14.ts
@@ -76,12 +76,9 @@ const card: Card = {
 
 	retreat: 2,
 
-	variants: {
-		normal: true,
-		reverse: false,
-		holo: false,
-		firstEdition: false
-	}
+	variants: [
+		{ type: "normal" }
+	]
 }
 
 export default card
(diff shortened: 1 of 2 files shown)


thomas-bassett (changes requested):
thanks for the PR, instead of changing the boolean values could you please update this card to use the new `variant_detailed` structure instead

santichausis:
Hi @thomas-bassett. Updated! Replaced the boolean variants with the variant_detailed array structure as requested.

thomas-bassett (approved):

KingTuttTuttle:
You guys killed it! Nice work and thank you!

PR #1372, opened 2026-04-11
fix(data): backfill set aliases and alias lookups

lamplis:
## Summary
- backfill visible abbreviations and source-backed `searchAliases` across remaining legacy, modern, promo, McDonald's, Trainer Kit, and TCG Pocket sets
- resolve set aliases consistently in direct set routes and card `set=` filters without exposing the internal alias list in public API payloads
- add supporting audit/research artifacts, unit tests, Bruno regressions, and `PORT` override support for isolated API validation

## Test plan
- [x] `bun run set-alias:audit`
- [x] `bun run validate`
- [x] `cd server && bun run --bun validate`
- [x] `cd server && bun test`
- [x] `cd .bruno && bru run --env Developpement --env-var BASE_URL=http://127.0.0.1:3322`

### DIFF
--- a/meta/definitions/api.d.ts
+++ b/meta/definitions/api.d.ts
@@ -57,11 +57,26 @@ interface variants {
 	wPromo?: boolean
 }
 
-interface variant_detailed {
+export interface variant_detailed {
 	type: string
 	size?: string
 	stamp?: Array<string>
 	foil?: string
+	thirdParty?: {
+		cardmarket?: number
+		tcgplayer?: number
+	}
+	image?: string
+}
+
+export interface SubsetCardCount {
+	official: number;
+}
+
+export interface SubsetInfo {
+	id: string;
+	name?: string;
+	cardCount: SubsetCardCount;
 }
 
 export interface SetResume {
@@ -79,13 +94,22 @@ export interface SetResume {
 		 */
 		official: number;
 	};
+	abbreviation?: {
+		official?: string;
+		localized?: string;
+	};
+	subsets?: Array<SubsetInfo>;
+	/**
+	 * Present on API payloads so GraphQL can resolve Set.serie on nested sets
+	 * (e.g. Card.set) without a separate field resolver.
+	 */
+	serie: SerieResume;
 }
 
 /**
  * /sets/:id
  */
 export interface Set extends SetResume {
-	serie: SerieResume;
 	tcgOnline?: string;
 	variants?: variants;
 	releaseDate: string;
@@ -137,7 +161,10 @@ export interface Set extends SetResume {
 		firstEd?: number;
 	};
 	cards: Array<CardResume>;
-	abbreviation: { official: string, localized: string };
+	abbreviation?: {
+		official?: string,
+		localized?: string
+	};
 	thirdParty?: {
 		cardmarket?: number
 		tcgplayer?: number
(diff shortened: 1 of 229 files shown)


meta/definitions/api.d.ts
```diff
@@ -57,11 +57,26 @@ interface variants {
 	wPromo?: boolean
 }
 
-interface variant_detailed {
+export interface variant_detailed {
 	type: string
 	size?: string
 	stamp?: Array<string>
 	foil?: string
+	thirdParty?: {
```
thomas-bassett, line 65: this is a change all by itself and is currently being done with PR1022

meta/definitions/api.d.ts
```diff
@@ -57,11 +57,26 @@ interface variants {
...
+		cardmarket?: number
+		tcgplayer?: number
+	}
+	image?: string
```
thomas-bassett, line 69: same as thirdparty

meta/definitions/api.d.ts
```diff
@@ -79,13 +94,22 @@ export interface SetResume {
 		 */
 		official: number;
 	};
+	abbreviation?: {
+		official?: string;
+		localized?: string;
+	};
+	subsets?: Array<SubsetInfo>;
```
thomas-bassett, line 101: handled by the V3 graphql. this required a change too the whole backend to do properly rather than a quick fix

meta/definitions/api.d.ts
```diff
@@ -79,13 +94,22 @@ export interface SetResume {
...
+	/**
+	 * Present on API payloads so GraphQL can resolve Set.serie on nested sets
+	 * (e.g. Card.set) without a separate field resolver.
+	 */
+	serie: SerieResume;
```
thomas-bassett, line 106: same as above `V3`

meta/definitions/api.d.ts
```diff
@@ -333,6 +360,28 @@ export interface Card extends CardResume {
 	 */
 	boosters?: Array<Booster>
 
+	/**
+	 * The card set number that appears on the card (e.g. 065/162, GG12/GG30)
+	 */
+	set_number: {
```
thomas-bassett, line 366: this is handled by the card count

server/compiler/utils/setUtil.ts
```diff
@@ -57,30 +93,39 @@ export async function getSetPictures(set: Set, lang: SupportedLanguages): Promis
 		const logoExists = file[lang]?.[set.serie.id]?.[set.id]?.logo ? `https://assets.tcgdex.net/${lang}/${set.serie.id}/${set.id}/logo` : undefined
 		const symbolExists = file.univ?.[set.serie.id]?.[set.id]?.symbol ? `https://assets.tcgdex.net/univ/${set.serie.id}/${set.id}/symbol` : undefined
 		return [
-			logoExists,
-			symbolExists
+			logoExists ?? set.logo,
```
thomas-bassett, line 96: what are we doing here, this was fine as it was. is there a specific issue this is addressing that i havnt seen?

server/compiler/utils/setUtil.ts
```diff
@@ -57,30 +93,39 @@ export async function getSetPictures(set: Set, lang: SupportedLanguages): Promis
...
 	const pics = await getSetPictures(set, lang)
+	const abbreviation = buildSetAbbreviation(set, lang)
+	const subsets = buildSetSubsets(set, lang)
+
 	return {
 		cardCount: {
-			official: set.cardCount.official,
-			total: Math.max(set.cardCount.official, cards.length)
+			official: set.cardCount?.official ?? 0,
+			total: Math.max(set.cardCount?.official ?? 0, cards.length)
```
thomas-bassett, line 113: null is better than 0. we will get questions asked and issues raised about inncorrect counts, let the user decide how to handle null counts

thomas-bassett (commented):
So this is screaming to me massivly that is was written predominatly/only by AI. While an improved search would be good this approach is not the way to go about it. really you would want to be looking into a "full text search" thats created automaticly at compile time. This PR is also kinda a mess with whats its adding, you expand the `variants_detailed` to have thirdparty and images which on its own is a single PR.

I belive "we" have spoken about AI prs before. Our stance is that it should not be used on its own espicailly when it comes to data as it has the tendancy to be wrong (ie the French abbreviation for ME02.5 is also `ASC` not `ME02.5` and `ME05` is not eternals. Please activly work on your changes as I think the majority of what you tried todo was already acomplishable with the current filtering

PR #1378, opened 2026-04-15
(data): Move thirdParty into Variants for MEG

FalconChipp:
## Changes

Moves all thirdParty IDs into the variants field as well as adding some missing variants for Mega Evolutions Base Set


thomas-bassett:
Conflicts! then will merge

thomas-bassett (dismissed):

PR #1384, opened 2026-04-17
[feat]: Add pull-rate types and derive rates

FalconChipp:
Add Phase 1 pull rate support
Introduces set-level pull rate data and derives it onto card variants at compile time.
New types (interfaces.d.ts, meta/definitions/api.ts)

PullRateValue — a display string or { display, percent } object for sortable/calculable rates
SpecialVariantPullRate — a variant match rule with an associated rate
PullRates — container for rarity rates and special variant rules

Data model changes

Set.pullRates — source of truth for set-level odds
Set.boosters[x].pullRates — optional per-booster override
Pull rates are authored on sets only, never on individual cards

Derivation (pull-rates.ts)

deriveVariantPullRate — matches a card variant against set.pullRates.specialVariants using explicit match fields; first match wins
Wired into cardUtil.ts — each variants_detailed entry gains a pullRate field at compile time

Output changes

set.pullRates passed through in setUtil.ts
variants_detailed[].pullRate populated on cards from sets with pull rate data


Pepper0ni:
How are you sourcing your pull rate data?

I have been doing my best to collect pull rate data for some time in order to power my Tabletop Simulator based Pack Simulator, and for ASC especially I have compiled an N of nearly 20000 for most slots, and the provided numbers don't come close to mine ( [quick reference](https://github.com/Pepper0ni/TTS-PTCG-Pack-Simulator/wiki/Pull-rate-and-card-order-Information), [source spreadsheets](https://github.com/Pepper0ni/TTS-PTCG-Pack-Simulator/blob/main/Pack%20rarity%20calculations%20and%20sources.ods) ), in particular your pull rate for MHRs is nearly twice mine (over 1/1000).

I also question the methodology here: we do not have official pull rates, and have not since the wizards era.  Every bit of data we have has to be collected from samples, and is thus subject to statistical uncertainty, and for older sets our data is very limited, often to rumors and forum heresay. The idea of an authoritative source for this is questionable, and having nothing to back it up makes it little better than the old forums. 

If I was to do this, I would collect and format the data differently. Instead of collecting rates in the API, I would store samples, and then have a script regularly (no reason to do this in real time) process those samples into usable rates with sample size provided for uncertainty calculations, with users being able to submit their recorded (as in video recording with the box being opened on stream, to avoid manipulation and scaling) samples to add to the store.

This is of course a great undertaking, but underlines that I don't think what is happening in this PR is good practice.

FalconChipp:
This PR is for adding pull rate fields, not for building an entire audited community statistics platform.

I’m not claiming pull rates are official or perfectly authoritative. Obviously modern physical pull rates are sample-based unless explicitly published. That is not some revelation, and it does not make the feature inherently bad practice.

If the Ascended Heroes figures are off compared to your data, then the useful part of that comment is the data difference. The rest of it is basically arguing for a completely different project and pretending that has to exist before this feature can exist at all, which I don’t agree with.

Also worth pointing out that the initial data in this PR was pulled together from several sources and a middle ground was chosen between them. It is not intended to represent final locked-in data, it is there to represent the shape and usage in the PR.

If you have stronger numbers and are willing to share the methodology/source, that is useful. But “I would architect a much bigger system for this” is not the same thing as this PR being bad practice and frankly outside the scope of what the api is.

Pepper0ni:
Sorry if I over-reacted, I get twitchy with data management and sources in a post-AI world, and suspected a hallucination in the data or vastly insufficient sample size as I have never recorded MHR rates that good after summing my sources.

As for sharing data, I am very happy to and already have via the links sent in my first comment. I typically collect my data from known pack opener posts and, in times of need, watching box opening streams and recording the results, what I have is in the spreadsheet linked. I've been at this a while and have been compiling sources since pull rate posts started becoming popular over covid, and have personally invested sets that did not have any data at all or who had poor data availability.

My suggestion for a new system is simply trying to make a data driven approach for what I consider to be subjective and thus challangable data. If you're not willing to do that, then at least posting what sources you do have would do a lot to address the underlying concerns.

FalconChipp:
Again, as stated in my last comment, the data used for the single set in this PR is not intended to be the final representation of the data. It is there to demonstrate the shape and usage of the field in the PR.

And as also already stated, a full community validation/reporting system with automated checks is far outside the scope of what this API is.

Issue #567, opened 2024-11-05
issue: invalid data returned when filtering using `set`


Aviortheking:
### What seems to be the issue

Card

### Field(s)

database

### Description

Discord message saying a request is not returning the correct datas

https://api.tcgdex.net/v2/en/cards?name=umbreon&set=evolving%20skies

Issue #566, opened 2024-10-30
issue: specific request crash the request


Aviortheking:
### Where are you having the issue

- [ ] GraphQL
- [X] JSON API

### Is the issue affecting one of the two elements below ?

- [ ] Translations
- [ ] Definitions Files (Open API, GraphQL Schema, etc)

### Description

https://api.tcgdex.net/v2/fr/cards?name=eq:Pikachu&name=eq:Pichu,Pikachu

Aviortheking:
The error
```
2024-11-05T17:56:46.563Z bruno-runtime/1.33.0             GET                   /v2/fr/cards?name=eq:Pikachu&name=eq:Pichu,Pikachu
-------------------------------------------------------------- ERROR --------------------------------------------------------------
127 | 			if (!query[key]) {
128 | 				query[key] = params
129 | 			} else {
130 | 				if (isObject(params)) {
131 | 					objectLoop(params, (v, k) => {
132 | 						(query[key] as any)[k] = v
                   ^
TypeError: Attempted to assign to readonly property.
      at /home/avior/Documents/tcgdex/cards-database/server/src/libs/QueryEngine/parsers.ts:132:14
      at objectLoop (/home/avior/Documents/tcgdex/cards-database/server/node_modules/@dzeio/object-util/dist/ObjectUtil.mjs:27:18)
      at /home/avior/Documents/tcgdex/cards-database/server/src/libs/QueryEngine/parsers.ts:131:6
      at objectLoop (/home/avior/Documents/tcgdex/cards-database/server/node_modules/@dzeio/object-util/dist/ObjectUtil.mjs:27:18)
      at recordToQuery (/home/avior/Documents/tcgdex/cards-database/server/src/libs/QueryEngine/parsers.ts:105:2)
      at /home/avior/Documents/tcgdex/cards-database/server/src/V2/endpoints/jsonEndpoints.ts:70:18
      at handle (/home/avior/Documents/tcgdex/cards-database/server/node_modules/express/lib/router/layer.js:95:5)
      at trim_prefix (/home/avior/Documents/tcgdex/cards-database/server/node_modules/express/lib/router/index.js:328:13)
      at /home/avior/Documents/tcgdex/cards-database/server/node_modules/express/lib/router/index.js:286:9
      at next (/home/avior/Documents/tcgdex/cards-database/server/node_modules/express/lib/router/index.js:280:10)
[...]

Issue #562, opened 2024-10-17
issue: GraphQL `set` request is not returning the correct value


Aviortheking:
### Where are you having the issue

- [X] GraphQL
- [ ] JSON API

### Is the issue affecting one of the two elements below ?

- [ ] Translations
- [ ] Definitions Files (Open API, GraphQL Schema, etc)

### Description

GraphQL `set` request does not respect the id filter
```gql
query {
	set(id: "sv07") {
		id,
		name
	}
}
```

Issue #489, opened 2024-05-18
issue: crash when sorting


Aviortheking:
### Where are you having the issue

- [ ] GraphQL
- [X] JSON API

### Is the issue affecting one of the two elements below ?

- [ ] Translations
- [ ] Definitions Files (Open API, GraphQL Schema, etc)

### Description

500 error when requesting `/v2/en/cards?name=pikachu&sort:field=hp&sort:order=DESC`

PR #1392, opened 2026-04-22
feat(Base Set): move thirdParty ids onto variants_detailed + additive schema/server retrocompat

lamplis:
## Summary

Migrate every card in `data/Base/Base Set/` (102 files) so that `thirdParty` (`cardmarket`, `tcgplayer`) lives on `variants_detailed[*]` instead of at the card root, plus the minimal additive schema widening and server-side reconstruction needed to keep both the V2 API response and the compiled JSON artifact byte-compatible for existing consumers.

## Files in this PR (107 total)

- Data: `data/Base/Base Set/*.ts` (102 files)
- Schema: `interfaces.d.ts`, `meta/definitions/api.d.ts`
- Server retrocompat: `server/src/V2/Components/Card.ts`, `server/compiler/utils/cardUtil.ts`, `server/compiler/utils/translationUtil.ts`

## Data changes

- Root-level `thirdParty` blocks are moved into the canonical `variants_detailed` entry for each Base Set card.
- Where multiple printings exist (1999-2000-copyright, shadowless, unlimited, 1st edition), Cardmarket product ids are attached to the specific variant they belong to.
- No other card fields are touched.

## Schema changes (additive only)

- `CardmarketVersion` token type (`V${number}` | `BASE` | `LIBRE`) added to `VariantType`. Used for placeholder variants emitted by the Cardmarket sync when a productId cannot be matched to a canonical source variant.
- `'Yellow A'` literal added to `VariantType` for alternate-art sibling printings that share a name and number with a base printing but use a local id suffixed with `"a"` (for example `XY150a`, `PHF024a`, `FFI055a`).
- Optional `variant_detailed.cardmarketLabels?: string[]` for the raw OCR-derived labels used by human reviewers to reconcile placeholder variants to canonical type / subtype / stamp.

## Server retrocompat

- Live V2 API (`server/src/V2/Components/Card.ts`):
  - `resolveRootPricingFromVariants` reconstructs root-level `cardmarket` / `tcgplayer` pricing from `variants_detailed[0].thirdParty`, falling back to the first variant that carries a populated pricing block.
  - Per-variant pricing is now populated from `variant.thirdParty`.
  - Legacy path preserved: cards that still hold root `thirdParty` (object-variant promos, trainer kits) continue to work unchanged.
- Compiled JSON (`server/compiler/utils/cardUtil.ts`):
  - `resolveRootThirdPartyFromVariants` synthesises the deprecated root `thirdParty` field from the first `variants_detailed` entry that carries one, so that direct JSON consumers keep seeing the same shape.
  - Root `thirdParty` is only `@deprecated` in the `Card` interface docstring; no behavior change for clients.
- `server/compiler/utils/translationUtil.ts`: small additive support for the new variant tokens (`CardmarketVersion`, `Yellow A`).

## Retrocompat guarantees

- V2 API: `card.pricing.cardmarket` / `card.pricing.tcgplayer` continue to resolve correctly for migrated cards via the new variant-based reconstruction. Non-migrated cards keep using the existing root `thirdParty` code path with zero behavior change for them.
- Compiled JSON: root `thirdParty` is preserved byte-for-byte on migrated cards through the new compile-time synthesis helper. Verified on `base1-1` (Alakazam) and `base1-10` (Mewtwo): same `{cardmarket, tcgplayer}` object as before the migration. Non-migrated `base2-1` (Clefable) is unchanged.
- GraphQL: no breaking change. The schema is derived from the same interfaces, and every addition is an optional field or widened union.

## Scope (intentional)

This PR is intentionally scoped to a single set so the variant move plus additive schema and server retrocompat shape can be reviewed end-to-end on a small, well-understood data surface (Base Set, all printings). Once merged, the remaining root-`thirdParty` cards can be moved in a single mechanical follow-up PR using the same canonical-variant landing rule, since the schema and server pieces only need to land once.

## Validation

Locally ran the full upstream CI gate (`.github/workflows/test.yml`), including Bruno integration:

| Gate | Result |
| --- | --- |
| `bun install --frozen-lockfile` (root + server) | clean |
| `bun run validate` (root TS, `tsc --noEmit`) | green |
| `cd server && bun run compile` | exit 0, 36236 / 36238 files loaded (the 2 `could not load file` warnings come from the pre-existing empty-path corner in `server/compiler/utils/util.ts` and are not introduced here) |
| `cd server && bun run --bun validate` (server TS, `tsgo --noEmit`) | green |
| `cd .bruno && bru run --env Developpement` | **68 / 68 requests, 121 / 121 assertions** - covers REST cards, filtering, random, sets, GraphQL (`card`, `cards`, `serie`, `series`, `set`, `sets`), and regression tickets #467, #471, #474, #475, #489, #562, #566, #567 |

Compiled JSON spot check (`server/generated/en/cards.json`):
- `base1-1` Alakazam (migrated): root `thirdParty = {cardmarket: 273696, tcgplayer: 42346}`, reconstructed from `variants_detailed[0].thirdParty`.
- `base1-10` Mewtwo (migrated): same pattern, retrocompat confirmed.
- `base2-1` Clefable (Base Set 2, not in this PR): unchanged, no regression on non-migrated cards.


lamplis:
Hi @thomas-bassett ,
what do you think about migrating set by set to the new variants model maintaining retro-compatibility?
When we validate this first commit, I'll send the other sets, one by one.

.github/scripts/load-cards.ts
```diff
@@ -10,6 +10,7 @@ type CardData = {
 	rarity?: string;
 	set: { name: string };
 	hasImage: boolean;
+	cardmarketIds?: number[];
```
FalconChipp: Can we remove all edits within .github. 
There isn't a reason to change the comments or actions for this

.github/workflows/comment-pr.yml
```diff
@@ -18,6 +18,7 @@ jobs:
         working-directory: .github/scripts
     permissions:
       pull-requests: write
+      contents: read
```
FalconChipp: Same here please

FalconChipp (changes requested):

data/Base/Base Set/1.ts
```diff
@@ -86,20 +86,27 @@ const card: Card = {
 	variants: [
 		{
 			type: "holo",
-			subtype: "unlimited",
+			subtype: "1999-2000-copyright",
+			thirdParty: {
+				cardmarket: 273696,
```
thomas-bassett, line 91: How are you sourcing this information? I checked just this one because it's very a specific subtype and its just linking to the standard

thomas-bassett (changes requested):

data/Base/Base Set/1.ts
lamplis: For me subtype: "1999-2000-copyright" is the basic normal card.
there is no "normal" variant so I mapped it to the standard cardmarket card.
But if "1999-2000-copyright" is not the basic original card, should I create a basic "normal" variant instead?

.github/scripts/load-cards.ts
lamplis: I added a link to cardmarket to test the cardmarket thirdparty id
If you don't want it, I'll remove it @FalconChipp

data/Base/Base Set/1.ts
lamplis: I found default basic card seem to be the "unlimited" (https://youtu.be/FIgZahAte7A?si=aY-stT7SDftmVQ4v)
Do you agree? @thomas-bassett

.github/scripts/load-cards.ts
FalconChipp: Please. Up to Aivor to change the git workflows.
If anything, put it in a separate pr and let him review when he gets the time

data/Base/Base Set/1.ts
thomas-bassett: the `1999-2000` copyright version is the 4th print and likely is just thrown in on cardmarket to be 1 id. 
but the "base" version of the card the way i looked at things should be the first print of the card.

in this card the `first edition shadowless` would be the first card printed

lamplis:
Hello,
Changes implemented :) @thomas-bassett @FalconChipp 
thx for review

lamplis:
The windows job often fails.
I proposed a fix in https://github.com/tcgdex/cards-database/pull/1370

FalconChipp (approved):
Git workflow changes removed

data/Base/Base Set/1.ts
```diff
@@ -86,16 +86,12 @@ const card: Card = {
 	variants: [
 		{
 			type: "holo",
-			subtype: "1999-2000-copyright",
```
thomas-bassett: why are you switching this around. I think you missed my point in the other comment. the base version "should" be the first print.

thomas-bassett (changes requested):
Please clean up the PR. the description still has all of the old changes in it. same with title. 

Also if we are going to move the ids over can get get all of the IDs so we can merge it once? Falcon has some tools that help with the extraction

lamplis:
Thanks Thomas. I cleaned up the PR title and description so they now match the actual 107-file diff.

On the broader "move all of the ids over once" point: I kept this PR scoped as the proof point for the full shape change, since it is the first one that combines (1) the additive `VariantType` widening, (2) `cardmarketLabels`, (3) the server-side retrocompat reconstruction, and (4) the first full-set migration on a set with multiple printings (Base Set 1st edition / shadowless / unlimited). That lets the schema/server contract be reviewed separately from a much larger mechanical data sweep.

That is also the same cadence Falcon has been using in the recent merged variant-migration PRs (`#1375`, `#1376`, `#1378`, `#1381`, `#1382`, `#1393`, `#1408`, `#1409`, `#1410`). Once this lands, I can follow immediately with one mechanical bulk PR for the remaining root-`thirdParty` cards so we only do the schema/server part once.

If you would rather I close this and reopen as one giant PR instead, say the word and I will do that. The tradeoff is just that the additive schema/server review then gets mixed into a much larger data-only migration.

server/compiler/utils/cardUtil.ts
```diff
@@ -168,10 +168,29 @@ export async function cardToCardSingle(localId: string, card: Card, lang: Suppor
 		})) : undefined,
 		updated: await getCardLastEdit(localId, card, lang),
 
-		thirdParty: card.thirdParty
+		thirdParty: card.thirdParty ?? resolveRootThirdPartyFromVariants(card.variants)
 	}
 }
 
+/**
+ * Retrocompat helper: reconstruct the deprecated root-level `thirdParty`
```
thomas-bassett, line 176: this already exsists just not at compile time

thomas-bassett:
> Thanks Thomas. I cleaned up the PR title and description so they now match the actual 107-file diff.
> 
> On the broader "move all of the ids over once" point: I kept this PR scoped as the proof point for the full shape change, since it is the first one that combines (1) the additive `VariantType` widening, (2) `cardmarketLabels`, (3) the server-side retrocompat reconstruction, and (4) the first full-set migration on a set with multiple printings (Base Set 1st edition / shadowless / unlimited). That lets the schema/server contract be reviewed separately from a much larger mechanical data sweep.
> 
> That is also the same cadence Falcon has been using in the recent merged variant-migration PRs (`#1375`, `#1376`, `#1378`, `#1381`, `#1382`, `#1393`, `#1408`, `#1409`, `#1410`). Once this lands, I can follow immediately with one mechanical bulk PR for the remaining root-`thirdParty` cards so we only do the schema/server part once.
> 
> If you would rather I close this and reopen as one giant PR instead, say the word and I will do that. The tradeoff is just that the additive schema/server review then gets mixed into a much larger data-only migration.

Did you? i still see all the server changes not needed in a Data PR. these MUST be split up unless abosulty nessasary like a new feature. 

I dont think the server changes you made are nessasary but its a mess im not sure what changes are what. this should be like 3 differnt prs (data,cmversion,thirdparty fallback) although the last one isnt needed. 

also looking at #1381 all variants have been given third-party ids. this is what I meant about filling out all the ids. If we don't do this now we will just need to come back and do them again later so lets get them done now.

I am going to close this PR for now. please split this up into small concise PRs and we can go from there.

PR #1394, opened 2026-04-24
Standardise the suffix of pokemon ex to "ex"

Pepper0ni:
## Changes

Suffix for pokemon ex (Not EX, they are different) is now reliably set to "ex"

## Details

Currently some pokemon ex have the same EX suffix as gen 5 and 6 pokemon EX's while most have nothing set. This PR consolidates it to reliably have the proper suffix of "ex".


data/EX/Crystal Guardians/89.ts
```diff
@@ -28,7 +28,7 @@ const card: Card = {
 		fr: "Galegon"
 	},
 
-	suffix: "EX",
+	suffix: "ex",
```
FalconChipp, line 31: Check the interfaces, "ex" should be capitalised

FalconChipp (dismissed):

data/EX/Crystal Guardians/89.ts
Pepper0ni: I believe this to be erroneous and misleading. 

EX and ex are considered different card types by the rules of the game, despite how similar the terms are. Cards that affect 1 do not affect the other, cards that evolve from an EX cannot evolve from their ex of the same species and they follow different mechanical frameworks (EX are always basic unless mega, even for evolved mons, ex match the stage of evolution their species is). 

Deliberately obscuring this information in the field that is best suited to identify the card's type is going to force users of the API who want to search by card type to use name checks to work around the overlap, rendering the field useless. Thus I would consider using "EX" to refer to exs a bug to be fixed, a bug made worse by just how easy it is to mix these 2 up thanks to the frankly confusing naming scheme. 

I will update the interfaces though for the new case.

thomas-bassett: This is fine adding the lower case `ex`

data/Scarlet & Violet/SVP Black Star Promos/204.ts
```diff
@@ -12,9 +12,13 @@ const card: Card = {
 		es: "Garchomp ex de Cintia",
 		pt: "Garchomp ex da Cíntia"
 	},
+<<<<<<< HEAD
```
thomas-bassett: might want to fix this

thomas-bassett (changes requested):

data/EX/Crystal Guardians/89.ts
FalconChipp: They also need adding to the translations and a comment adding to document the difference

Pepper0ni: Do you want the comment adding to the interfaces or somewhere else?

thomas-bassett (approved):

PR #1395, opened 2026-04-25
data: declare reverse holo variants for SV03.5 set

mathieu-gilloots:
Common, Uncommon and Rare cards in the Scarlet & Violet 151 set all have a Reverse Holofoil print available (visible on Cardmarket, TCGplayer, and Bulbapedia) but the source files only declared the variants they explicitly disabled. The API thus returned `reverse: false` for all 153 cards in scope, hiding the existence of these prints to consumers.

This commit makes the variants explicit:
- Common (66 cards): { normal: true, reverse: true }
- Uncommon (62 cards): { normal: true, reverse: true }
- Rare (25 cards): { normal: false, holo: true, reverse: true } (Rares in SV-era are holo by default, no non-holo print exists)

Other rarities (Ultra/Double/Illustration/Special illustration/Hyper) left untouched, since they only have their special print.

## Details

151 cards have their reverse variant.

### DIFF
--- a/data/Scarlet & Violet/151/001.ts
+++ b/data/Scarlet & Violet/151/001.ts
@@ -47,9 +47,14 @@ const card: Card = {
 	retreat: 2,
 	regulationMark: "G",
 
-	variants: {
-		holo: false
-	},
+	variants: [
+		{
+			type: "normal"
+		},
+		{
+			type: "reverse"
+		}
+	],
 
 	illustrator: "Yuu Nishida",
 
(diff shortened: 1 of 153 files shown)


thomas-bassett (changes requested):
Hey @mathieu-gilloots thanks for the pr. We are going through and updating these manually at the moment. Would you mind please using the new variants_detailed format instead of the boolean?

mathieu-gilloots:
@thomas-bassett like that ?

thomas-bassett (approved):
That will do nicely will save us a little bit of work

PR #1414, opened 2026-05-03
Fix M3 Card count to 80

SierraOne4:
## Changes

Corrects card count to 80.

### DIFF
--- a/data-asia/M/M3.ts
+++ b/data-asia/M/M3.ts
@@ -10,7 +10,7 @@ const set: Set = {
 	serie: serie,
 
 	cardCount: {
-		official: 117,
+		official: 80,
 	},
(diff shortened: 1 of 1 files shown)


data-asia/M/M3.ts
```diff
@@ -10,7 +10,7 @@ const set: Set = {
 	serie: serie,
 
 	cardCount: {
-		official: 174,
+		official: 117,
```
thomas-bassett: both of these are wrong. official count should be 80

thomas-bassett (changes requested):

SierraOne4:
@thomas-bassett My bad! Fixed now.

thomas-bassett (approved):

PR #1421, opened 2026-05-06
(data): Move thirdParty into variants for Base Set

lamplis:
## Summary
- move Base Set root `thirdParty` ids onto existing `variants` entries only
- keep the change pure data: no schema, server, workflow, or tooling files
- align generic marketplace ids with the intended Base Set print order and keep clear jumbo listings with existing variant fields

## Test plan
- [x] `bun run validate`
- [x] `cd server && bun run --bun validate`
- [x] `cd .bruno && bru run --env Developpement`

### DIFF
--- a/data/Base/Base Set/1.ts
+++ b/data/Base/Base Set/1.ts
@@ -86,7 +86,15 @@ const card: Card = {
 	variants: [
 		{
 			type: "holo",
-			subtype: "unlimited",
+			subtype: "1999-2000-copyright",
+			thirdParty: {
+				cardmarket: 273696,
+				tcgplayer: 42346
+			}
+		},
+		{
+			type: "holo",
+			subtype: "unlimited"
 		},
 		{
 			type: "holo",
@@ -96,10 +104,9 @@ const card: Card = {
 		{
 			type: "holo",
 			subtype: "shadowless",
-		},
-		{
-			type: "holo",
-			subtype: "1999-2000-copyright",
+			thirdParty: {
+				cardmarket: 660227
+			}
 		}
 	],
 
@@ -108,10 +115,6 @@ const card: Card = {
 		it: "Il suo cervello è più potente di un supercomputer. Dicono che abbia un quoziente di intelligenza di 5.000. LIV 42 N.65"
 	},
 
-	thirdParty: {
-		cardmarket: 273696,
-		tcgplayer: 42346
-	}
 }
 
 export default card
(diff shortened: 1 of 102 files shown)


lamplis:
Hi,
Reading the following article, I understand we are missing variants : https://relentlessdragon.com/pokemon-card-game/identifying-early-pokemon-cards/#:~:text=In%20fact%2C%20the%20first%20print,than%20an%20unlimited%20edition%20card.

I will propose a new mapping of card variants and cardmarket’s versions

There should be 2 shadowless variants for example(1st edition and Limited)

thomas-bassett:
> Hi, Reading the following article, I understand we are missing variants : https://relentlessdragon.com/pokemon-card-game/identifying-early-pokemon-cards/#:~:text=In%20fact%2C%20the%20first%20print,than%20an%20unlimited%20edition%20card.
> 
> I will propose a new mapping of card variants and cardmarket’s versions
> 
> There should be 2 shadowless variants for example(1st edition and Limited)

im unable to check as cloudflare blocks me 🗡️  but we should have those already it might just be naming convention

type: "holo",
subtype: "unlimited",

type: "holo",
subtype: "shadowless",
stamp: ["1st-edition"]

type: "holo",
subtype: "shadowless",

as far as im aware and i cant check your link but all the unlimiteds/limited where shadowed. so the 2 shadowless variants are covered by variant 2,3

data/Base/Base Set/1.ts
```diff
@@ -86,20 +86,27 @@ const card: Card = {
 	variants: [
 		{
 			type: "holo",
```
thomas-bassett, line 88: why are we messing with the order of variants? apart from moving the 1st edition print as disccused before there shouldnt be a need for any other change

data/Base/Base Set/1.ts
```diff
@@ -86,20 +86,27 @@ const card: Card = {
...
-			subtype: "unlimited",
+			subtype: "1999-2000-copyright",
 		},
 		{
 			type: "holo",
-			subtype: "shadowless",
-			stamp: ["1st-edition"]
+			subtype: "unlimited"
```
thomas-bassett, line 93: still missing thirdparty ids. there might be some obvious ones that wont have ids but holo-unlimited definatly should have one

thomas-bassett (changes requested):
Make sure all have third-party ids, theres potentially a missprint variant in here that wont have that but nearly all should have them.

also your shuffling the ordering around without much reason to. if you want to change the order the shadowless-1stedition is the first card printed and is likely the most logical 1st variant

advra:
i dont think this is needed anymore. Work already completed with https://github.com/tcgdex/cards-database/pull/1686

MineTommy:
Yes

Also PR #2205 added the Rest Cardmarket Ids

PR #1422, opened 2026-05-09
data(sv4m): add Korean texts for full set

LILQK:
## Summary\n- add Korean (ko) card texts for the full SV4M set in data-asia/SV/SV4M\n- update names/effects/descriptions across all 95 cards where applicable\n- keep existing languages intact and only add/adjust ko entries\n\n## Validation\n- bun run validate\n\n## Notes\n- source text extracted from official Korean card pages and mapped to SV4M card numbers\n- corrected a 091/092 alignment mismatch to match existing ja/th card content


server/src/V2/Components/Card.ts
```diff
@@ -50,7 +50,7 @@ pl.forEach((it) => list[`${it.id.toLowerCase()}pl`] = it)
 // @ts-expect-error there is currently not cards here
 ru.forEach((it) => list[`${it.id.toLowerCase()}ru`] = it)
 ja.forEach((it) => list[`${it.id.toLowerCase()}ja`] = it)
-// @ts-expect-error there is currently not cards here
+// @ts-ignore ts can't load file
```
thomas-bassett: lets remove this for now. i know it fixes but this is a data PR

thomas-bassett (changes requested):
just need to remove the non data change

thomas-bassett (approved):

PR #1429, opened 2026-05-12
fix: update EN data for Base Set 1

philgruneich:
## Changes

Update dataset for Base Set 1 in English


data/Base/Base Set/96.ts
```diff
@@ -16,6 +16,7 @@ const card: Card = {
 	energyType: "Special",
 
 	effect: {
+		en: "Provides ColorlessColorless energy.\nDoesn't count as a basic Energy card.",
```
philgruneich: I'd like a double check for this one in case the formatting should be different, like avoiding `\n` and if we should replace Colorless for the pip abbreviation between curly brackets, etc.

thomas-bassett: `\n` is fine its used in other cards

the Colorless depends on how its written on the card. if its the word then it stays the word if its the symbol its `{C}`

thomas-bassett (changes requested):

thomas-bassett (approved):

PR #1435, opened 2026-05-12
feat(data-asia): add Japanese set M2a (MEGAドリームex)

LILQK:
Adds the Japanese **M2a** high-class pack **MEGAドリームex**: set metadata in \data-asia/M/M2a.ts\ and **250** card files \


thomas-bassett (changes requested):
Lets check ALL variants, they cant all be holo

LILQK:
I’m fixing a few things and looking for a source to verify the correct variants.
This commit can be ignored.

thomas-bassett:
closing due to inactivity

PR #1446, opened 2026-05-13
patch: EN data E-Card/Aquapolis

philgruneich:
## Changes

Updates EN data for E-Card/Aquapolis


data/E-Card/Aquapolis/118.ts
```diff
@@ -14,6 +14,7 @@ const card: Card = {
 	set: Set,
 
 	effect: {
+		en: "This card stays in play when you play it. Discard this card if another Stadium card comes into play. Once during each player's turn (before attacking), if that player's Bench isn't full, that player flips a coin. If heads, that player shows his or her opponent a basic Energy card from his or her hand. Then, that player searches his or her deck for a Basic Pokémon card of the same type (color) as the revealed Energy card and puts it onto his or her Bench. The player shuffles his or her deck afterward.",
```
philgruneich, line 17: There are some design decisions to be made here. See how I'm including the "Stadium" text (and other cards also have the "Supporter" text). The French translations in this file mostly contain only the "Stadium/Supporter text", and the German translation here is just not German.

data/E-Card/Aquapolis/122.ts
```diff
@@ -14,6 +14,7 @@ const card: Card = {
 	set: Set,
 
 	effect: {
+		en: "Attach this card to 1 of your Fire Pokémon in play. That Pokémon may use this card's attack instead of its own. At the end of your turn, discard Fire Cube 01.",
 		fr: "Attachez cette carte à l'un de vos Pokémon  en jeu. Ce Pokémon peut utiliser l'attaque de cette carte à la place de la sienne. À la fin de votre tour, défaussez-vous de Cube de feu 01.",
```
philgruneich, line 18: Notice how here and in the previous card we have the double space in the FR translation because it is expecting a pip or something else. I think that is also the issue in the DE one.

data/E-Card/Aquapolis/131.ts
```diff
@@ -14,6 +14,7 @@ const card: Card = {
 	set: Set,
 
 	effect: {
+		en: "This card stays in play when you play it. Discard this card if another Stadium card comes into play. Once during each of his or her turns, whenever a player attaches an Energy card from his or her hand to 1 of his or her Benched Pokémon, he or she removes 1 damage counter, if any, from that Pokémon.",
 		fr: "Cette carte reste en jeu lorsque vous la jouez. Défaussez -vous de cette carte si une autre carte Stade arrive en jeu.",
```
philgruneich, line 18: This is one of those cases where the FR translation only has the Stadium text and nothing else.

PR #1452, opened 2026-05-14
patch: EN data EX/Dragon

philgruneich:
## Changes

Updates EN data for EX/Dragon

### DIFF
--- a/data/EX/Dragon/82.ts
+++ b/data/EX/Dragon/82.ts
@@ -15,6 +15,7 @@ const card: Card = {
 	trainerType: "Tool",
 
 	effect: {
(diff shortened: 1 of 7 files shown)


data/EX/Dragon/82.ts
```diff
@@ -15,6 +15,7 @@ const card: Card = {
 	trainerType: "Tool",
 
 	effect: {
+		en: "Attach Balloon Berry to 1 of your Pokémon that doesn't already have a Pokémon Tool attached to it. If that Pokémon is Knocked Out, discard this card. When the Pokémon Balloon Berry is attached to retreats, discard Balloon Berry instead of discarding Energy cards.",
```
thomas-bassett, line 18: Can you let me know what the source is for this information? are you taking it direct from the card itself?

philgruneich: https://github.com/PokemonTCG/pokemon-tcg-data

philgruneich: I'm also doing random checks on the actual cards in different services.

PR #1560, opened 2026-05-25
feat: migrate Chaos Rising (MEP) to new variant format

pragmaticivan:
## Summary
- Moves `thirdParty` tcgplayer IDs into the `variants` array for all 122 cards
- Removes old boolean `variants` object fields (`normal`, `reverse`, `holo`, `firstEdition`)


FalconChipp:
Closed as duplicate

PR #1570, opened 2026-05-28
Add missing cards for SM Trainer Kit Alolan Raichu

0xHellCat:
I don't know why, but almost all the "energy" cards had not been added.


thomas-bassett (changes requested):
Thanks for your first PR couple things:

-  can you add the variants to these cards.
- please use the `001` format for the card naming

thomas-bassett:
closing due to inactivity

PR #1571, opened 2026-05-29
Fix/cardmarket ids

Cherden:
## Changes

- update cardmarket ids for cards in multiple sets

## Details

*This PR is a draft because its ongoing, I'm not finished yet*

I wanted to have a unique relation between cards and cardmarket ids and upon ingesting data noticed there are a lot of duplicates. This PR will fix the ones I found through my duplicate check script. The fix is mostly checking by hand, for some I was able to fix with a script (copy paste error from the previous card).

The behavior I noticed with the duplicate errors is that a previous script was grabbing the first cardmarket id it could find for a specific name and applying it to all card instances.

In other PRs I read that there is a new schema for variants which include external ids. Please note that this is out of scope for this PR, I'm only fixing the cardmarket ids.

Validation:

- pnpm validate


thomas-bassett:
Hi, thanks for the PR

Please check with the contributing guidelines, this is a massive PR. and needs splitting down into separate parts. Ideally, you wouldn't submit more than one set at once but if it's only a couple of cards, then you can combine them. 

Also, I see you've said that variants are out of scope. Just note that we may decide not to merge some PRs in favour of waiting for the variant version as its easier to manage that way

Cherden:
> Hi, thanks for the PR
> 
> Please check with the contributing guidelines, this is a massive PR. and needs splitting down into separate parts. Ideally, you wouldn't submit more than one set at once but if it's only a couple of cards, then you can combine them.
> 
> Also, I see you've said that variants are out of scope. Just note that we may decide not to merge some PRs in favour of waiting for the variant version as its easier to manage that way

Sure, can split it. Will wait until I get a signal when the variant migration is done, then.

PR #1609, opened 2026-06-05
Fix: add stamp types to interface for 'fernando-cifuentes' and 'fernando-cifuentes-gold'

jshill103:
closes #

### DIFF
--- a/interfaces.d.ts
+++ b/interfaces.d.ts
@@ -34,7 +34,8 @@ export type VariantStamps = '1st-edition' | 'w-promo' | 'pre-release' | 'pokemon
 	| 'zachary-bokhari' | 'shuto-itagaki' | 'snowflake' | 'trick-or-trade' | 'horizons' | 'gamestop' | 'eb-games'
 	| 'illustration-contest-2024' | 'worlds-2025' | 'top-eight' | "champion" | "master-ball-league" | "ultra-ball-league" | "judge" | "asia-promo"
 	| "international-championship-europe" | "international-championship-latin-america" | "international-championship-north-america" | 'ace-trainer'
-	| 'pikachu' | 'bulbasaur' | 'squirtle' | 'charmander' | 'pokeball' | '30th-pokeday' | 'mcdonalds' | 'pokemon-together'
+	| 'pikachu' | 'bulbasaur' | 'squirtle' | 'charmander' | 'pokeball' | '30th-pokeday' | 'mcdonalds' | 'pokemon-together' | 'fernando-cifuentes'
+	| 'fernando-cifuentes-gold'
 
 export interface variant_detailed {
 	/**


interfaces.d.ts
```diff
@@ -34,7 +34,8 @@ export type VariantStamps = '1st-edition' | 'w-promo' | 'pre-release' | 'pokemon
 	| 'zachary-bokhari' | 'shuto-itagaki' | 'snowflake' | 'trick-or-trade' | 'horizons' | 'gamestop' | 'eb-games'
 	| 'illustration-contest-2024' | 'worlds-2025' | 'top-eight' | "champion" | "master-ball-league" | "ultra-ball-league" | "judge" | "asia-promo"
 	| "international-championship-europe" | "international-championship-latin-america" | "international-championship-north-america" | 'ace-trainer'
-	| 'pikachu' | 'bulbasaur' | 'squirtle' | 'charmander' | 'pokeball' | '30th-pokeday' | 'mcdonalds' | 'pokemon-together'
+	| 'pikachu' | 'bulbasaur' | 'squirtle' | 'charmander' | 'pokeball' | '30th-pokeday' | 'mcdonalds' | 'pokemon-together' | 'fernando-cifuentes'
```
advra, line 37: Before we get too far in the weeds, shouldnt we instead have World Championship as a `VariantType`?

Then in variant_detailed subtypes define ` fernando-cifuentes`? It seems like drilling down these championship variants is becoming messy. It is easier to drill down through "Championship variants" first rather than the championship winner's specific names.

jshill103: I would agree but following what is already here in this interface: | 'zachary-bokhari' | 'shuto-itagaki' would follow the pattern im attempting to implement here. Tho I do think eventually something to what you described should be implemented.

thomas-bassett: so this is something i touched on earlier. i believe the PR is open still i will go back and update it. but i basically create a `stamp_detailed` object that would take a stamp say for example `world-championships` and then you would fill in the `detail` field with `2018` the  compiler at runtime would flatten the stamp into `world-championships-2018` this cleared up a large amount of these very similar stamps. 

i get what your saying about a variant type but i would argue what makes a card a world championship variant type, in my eyes its the stamp so moving world championships out of stamps removes some of the specific information about the card. How would a user know that all world championship cards have a stamp.

thomas-bassett:
only add the stamps when you use them, going to close this for now.

interfaces.d.ts
advra: I see. I agree with the `stamp_detailed` approach if it can clean up the variants and flatten then we will be at a much better state than previously. It would be a big improvement. In `detail` we can then  include the champions name somehow.

thomas-bassett: I would leave the base signatures in stamps so `zachary-bokhari` would stay in the stamps. But the `gold` and `silver` details wouldn't need too. This is to keep as much as possible type defined

PR #1613, opened 2026-06-09
editor: Added Reverse

TCGdexBOT:
Added Reverse


thomas-bassett:
These got moved to the new structure in the time this was opened

PR #1635, opened 2026-06-14
patch(data): ja fixed duplicate xy5a

advra:
closes # N/A

## Changes

- fixed duplicate XY5a (XY5b file setid was wrong so I changed it from XY5a to XY5b)
- added english translations for XY5a and XY5b

## Details

Japanese sets like [Gaia Volcano • Tidal Storm](https://bulbapedia.bulbagarden.net/wiki/Gaia_Volcano/Tidal_Storm_(TCG)) are split into two. These are the equivalent of Primal Clash. When doing so the set id for japanese should be [set]a and [set]b. It seems like someone did a copy paste error resulting in duplicate XY5a.


thomas-bassett (changes requested):
remove `en`

advra:
Done!

thomas-bassett (approved):

PR #1650, opened 2026-06-18
patch(update): ja corrections for SM2+ 

advra:
closes # NA

## Changes

- SM2+ setid
- apply correct translation Source: https://bulbapedia.bulbagarden.net/wiki/Facing_a_New_Trial_(TCG)

## Details

Wrong capitalization. Should be `SM2+` not `sm2+` 🥶


thomas-bassett:
for most things i like to go with what is written on the Cards and Booster boxes. in this case both have the lowercase id.

advra:
Ok thanks for confirmation! I truly prefer whats on the card as well but wasnt sure from my last changes. Thanks for confirming!

PR #1651, opened 2026-06-18
patch(update): ja corrections for CS4.5

advra:
closes # N/A

## Changes

- changed set id in CS4.5 

## Details

Should be CS4.5 not CS4


thomas-bassett (changes requested):
similar to the one i just closed can we correct the id to lowercase to match with the card and booster printings?

thomas-bassett:
closing due to inactivity

PR #1662, opened 2026-06-23
move thirdparty into variants and add some missing variants for me03

FatalMistake02:
## Changes

move thirdparty into variants and add some missing variants for me03

### DIFF
--- a/data/Mega Evolution/Perfect Order/001.ts
+++ b/data/Mega Evolution/Perfect Order/001.ts
@@ -60,17 +60,17 @@ const card: Card = {
 
 	variants: [
 		{
-			type: "normal"
+			type: "normal",
+			thirdParty: {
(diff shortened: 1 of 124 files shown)


data/Mega Evolution/Perfect Order/001.ts
```diff
@@ -60,17 +60,17 @@ const card: Card = {
...
 		{
-			type: "normal"
+			type: "normal",
+			thirdParty: {
+				tcgplayer: 684397,
+				cardmarket: 877413
+			}
 		},
 		{
 			type: "reverse"
```
FalconChipp: Please make sure the reverses have the same IDs please

FalconChipp (changes requested):

FalconChipp (approved):

PR #1701, opened 2026-07-05
fix: add detailed variants to pop 8 series cards

FelipePatitucci:
closes # 

## Changes

Added detailed variants for POP  8 series cards.

### DIFF
--- a/data/POP/POP Series 8/1.ts
+++ b/data/POP/POP Series 8/1.ts
@@ -69,12 +69,15 @@ const card: Card = {
 
 	retreat: 3,
 
-	variants: {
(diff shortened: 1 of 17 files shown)


FalconChipp:
Would you also be ok with adding the pricing IDs to the variants as part of this PR? 
If needed, please reach out and I can provide the JSON files containing the IDs

FelipePatitucci:
> Would you also be ok with adding the pricing IDs to the variants as part of this PR? If needed, please reach out and I can provide the JSON files containing the IDs

Sure, I will tag you on the discord server.

PR #1704, opened 2026-07-06
fix: point Brilliant Stars Trainer Gallery cards at their own set

Sinrath:
## Changes

Updates the `Set` import in the 30 Brilliant Stars Trainer Gallery card files from `../Brilliant Stars` to `../Brilliant Stars Trainer Gallery`.

## Details

#1693 moved the TG cards into their own set (`swsh9.5tg`) but kept the card files importing the parent set. The compiled output therefore lists the Trainer Gallery cards with their old `swsh9-TG*` IDs (visible on `/v2/en/sets/swsh9.5tg`), while `swsh9` no longer contains them.


FalconChipp:
Great catch, thank you

PR #1718, opened 2026-07-11
RFC + POC: @tcgdex/schema — single-source-of-truth output types [V3 - Not for Merge] 

FalconChipp:
> [!WARNING]
> **RFC + POC — not for merge as-is.** This PR models the V2 output shape end-to-end to prove the pipeline; the real package will model V3's output contract. See `RFC.md` for the full proposal.

> **Read the RFC as rendered markdown:** [`schema/RFC.md`](https://github.com/tcgdex/cards-database/blob/SDK-Compat/schema/RFC.md)
>
> The diff view mangles tables, code blocks, and the ASCII diagrams. This link renders it properly.

## Changes

Adds a new `schema/` package proposing TypeBox-defined output shapes as the single source of truth for SDK types and OpenAPI. 
Runnable POC that models the current V2 output shape end-to-end (types, OpenAPI, live validation) as the foundation for V3's output contract.

## Details

 everything lives under `schema/`.

- `RFC.md` — full proposal: problem (silent drift between input, compiled, and output shapes; SDK types every enum as `string`; `variants_detailed`/`pricing`/`updated` are served but never typed), solution (`@tcgdex/schema` with TypeBox as JSON-Schema-native source of truth), open-vs-closed enum policy, staging plan, FAQ
- `README.md` — how to run the demo, what to look at
- `src/` — the schema definitions (`Card`, `CardResume`, `Set`, `SetResume`, `Serie`, `SerieResume`, `Variants`)
- `build/emit-dts.ts` — generates `dist/types.d.ts` (closed enums become real unions; open enums stay `string` and carry known values as `x-extensible-enum` OpenAPI metadata)
- `build/emit-openapi.ts` — generates `dist/openapi.json` (full OpenAPI 3.1 doc: 33 paths, 9 component schemas, RFC 9457 `application/problem+json` errors)
- `build/validate-live.ts` — fetches `base1-4` from `api.tcgdex.net` and validates against the schema
- `docus/sdk-migration.md` — how the JS SDK adopts the schema via `implements` (types-only devDep, no runtime footprint)
- `docus/workflows/` — example GitHub Actions showing how schema releases would auto-regenerate downstream SDKs (illustrative, not  wired up)
- `docus/python-template-sketch.mustache` — sketch of the `x-extensible-enum` → real-enum Python override (Stage 2b work, non-blocking)

**Run it:**

```bash
cd schema && npm install && npm run demo
```

## Docs index

Quick navigation for reviewers — all rendered at the branch head:

**Core:**
- 📘 [`RFC.md`](https://github.com/tcgdex/cards-database/blob/SDK-Compat/schema/RFC.md) — the full proposal
- 📖 [`README.md`](https://github.com/tcgdex/cards-database/blob/SDK-Compat/schema/README.md) — how to run the demo, expected output

**Under `docus/`:**
- 🔧 [`sdk-migration.md`](https://github.com/tcgdex/cards-database/blob/SDK-Compat/schema/docus/sdk-migration.md) — JS SDK adoption via `implements`
- 📋 [`implementation-followups.md`](https://github.com/tcgdex/cards-database/blob/SDK-Compat/schema/docus/implementation-followups.md) — known POC gaps + RFC wording corrections to address
- ⚙️ [`workflows/README.md`](https://github.com/tcgdex/cards-database/blob/SDK-Compat/schema/docus/workflows/README.md) — example release → auto-regen loop
- 🐍 [`python-template-sketch.mustache`](https://github.com/tcgdex/cards-database/blob/SDK-Compat/schema/docus/python-template-sketch.mustache) — sketch of the `x-extensible-enum` Python override

**RFC section deep links (jump straight to a topic):**
- [Problem](https://github.com/tcgdex/cards-database/blob/SDK-Compat/schema/RFC.md#problem) — drift diagnosis + evidence
- [Language coverage](https://github.com/tcgdex/cards-database/blob/SDK-Compat/schema/RFC.md#language-coverage) — how any language consumes this
- [Public API contract](https://github.com/tcgdex/cards-database/blob/SDK-Compat/schema/RFC.md#the-schema-as-a-public-api-contract) — third-party use cases
- [Documentation site](https://github.com/tcgdex/cards-database/blob/SDK-Compat/schema/RFC.md#documentation-site) — how docs stop drifting
- [Considered alternatives](https://github.com/tcgdex/cards-database/blob/SDK-Compat/schema/RFC.md#considered-alternatives) — `typescript-json-schema` + shared enum hybrid
- [Staging](https://github.com/tcgdex/cards-database/blob/SDK-Compat/schema/RFC.md#staging) — 4-stage rollout
- [FAQ](https://github.com/tcgdex/cards-database/blob/SDK-Compat/schema/RFC.md#faq)

---

### DIFF
--- a/schema/.gitignore
+++ b/schema/.gitignore
@@ -1,0 +1,2 @@
+node_modules/
+dist/
--- a/schema/README.md
+++ b/schema/README.md
@@ -1,0 +1,71 @@
+# @tcgdex/schema — demo
+
+Runnable proof for the schema-driven-types RFC, modeled on the **current V2** output
+shape so it can be checked against live data and the existing `@tcgdex/sdk` types.
+The real package will define V3's output contract using the same pipeline.
+
+> **Not for merge as-is.** This POC mirrors V2 to prove the pipeline end-to-end.
+> See [`RFC.md`](./RFC.md) for the full proposal and staging plan.
+
+## Run
+
+    npm install
+    npm run demo      # emit:dts + emit:openapi + validate:live
+
+Expected output:
+
+    → dist/types.d.ts
+    ✓ wrote dist/openapi.json
+      33 paths, 9 schemas
+    ✓ live base1-4 validates against Card
+
+## What to look at
+
+1. **`dist/types.d.ts`** — clean TypeScript declarations generated from the schema.
+   Closed enums (`category`, `types`, `energyType`, `abilityType`) become real unions
+   like `"Pokemon" | "Trainer" | "Energy"`. Open enums (`rarity`, `stage`, `suffix`)
+   stay `string` — new values from future set releases won't break consumers.
+   Compare with the current `@tcgdex/sdk` `.d.ts` where every enum is bare `string`.
+
+2. **`dist/openapi.json`** — full OpenAPI 3.1 document: 33 paths, 9 component schemas,
+   RFC 9457 `application/problem+json` error responses, `x-extensible-enum` metadata
+   on every open enum. Feed this straight into `openapi-generator-cli` for non-JS SDKs.
+   Compare with the hand-maintained `server/public/v2/openapi.yaml` (1705 lines).
+
+3. **`npm run validate:live`** — fetches `base1-4` (Charizard, Base Set) from
+   `api.tcgdex.net` and validates it against the schema. On the current run, it
+   passes clean with no un-modeled fields — `variants_detailed`, `pricing`, and
+   `updated` (all ghost properties on the current SDK type) are now modeled.
+
+## Sample generated SDK output
+
+To prove the OpenAPI pipeline is real, not aspirational:
+
+```powershell
+npx @openapitools/openapi-generator-cli generate -i dist/openapi.json -g python -o dist/generated-python-sample --additional-properties=packageName=tcgdex
+```
+
+Requires Java (the generator is a JVM tool). Produces a full Python client with
+Pydantic models — see `dist/generated-python-sample/tcgdex/models/card.py`. The
+pipeline works end-to-end; the follow-up work described in the RFC is a Mustache
+template override to convert `x-extensible-enum` metadata into real Python enums
+with an `UNKNOWN` fallback for open-enum semantics.
+
+## Docs
+
+- [`RFC.md`](./RFC.md) — the full RFC (problem, solution, staging, FAQ)
+- [`docus/sdk-migration.md`](./docus/sdk-migration.md) — how the JS SDK adopts the schema (`implements` pattern)
+- [`docus/workflows/`](./docus/workflows/) — example GitHub Actions showing how schema releases would auto-regenerate downstream SDKs
+- [`docus/python-template-sketch.mustache`](./docus/python-template-sketch.mustache) — a sketch of the `x-extensible-enum` Mustache override (needs archaeology against the actual openapi-generator Python defaults before it will run)
+
+## Notes / production deltas
+
+- **Sub-schemas are inlined** in the generated OpenAPI (no `$ref`s between component
+  schemas). Reviewers will see auto-named inline schemas like `Card_variants_detailed_inner`
+  in the Python sample — cosmetic, not correctness. Production adds a `$ref`-lifting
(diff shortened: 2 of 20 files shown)


FalconChipp:
Also a possibility of using the interfaces as the source of truth and dropping TypeBox in favour of a `typescript to json` package to derive JSON schema from existing types.

Flips the dependency direction to `cards-database -> schema -> SDKs` which would follow the actual data flow and kills enum drift risk fully — and already recognisable to contributors.

However, this would mean that we lose the free runtime validation (not breaking, but worth pointing) and extensions like `x-extensible-enum` become JSDoc annotations rather than schema options, therefore tool-dependent. Also adds an extra TS compiler pass to the build where some TS constructs like template literals don't always convert cleanly.

PinKushin:
Hi — I maintain the C#/.NET SDK for TCGdex ([TcgDex.CSharpSdk](https://github.com/PinKushin/TcgDex.CSharpSdk), [NuGet](https://www.nuget.org/packages/TcgDex.CSharpSdk)). Ryan suggested I bring over what I've picked up building it, since it's effectively a third-party test of the problem this RFC describes: I model the same output shapes you do, but from outside the JS ecosystem, in a language where type mismatches are hard failures.

**Short version: I hit the drift you're diagnosing, independently, and a published schema would have caught every instance at build time.**

### Confirming the diagnosis

My SDK types `variants_detailed`, `pricing` and `updated` — the three fields your `validate-live` script flags as un-modeled — but only because I built my models from live responses instead of documentation. There was no way to know from any published contract that they were supposed to be there. That's independent confirmation that the drift is real and that the current source of truth is "curl and look."

### A live example of the schema/data mismatch, on your GraphQL endpoint today

```graphql
{ cards(filters: { rarity: "Common" }) { ... attacks { name } } }
```

fails with:

```
Cannot return null for non-nullable field AttacksListItem.name.
```

Some cards have attacks without names, but the schema declares `name` non-nullable — and GraphQL can't return a partial list for a non-nullable field, so the *entire query* errors rather than omitting one card. A narrow filter that happens to miss those cards succeeds, so it looks intermittent; it's actually deterministic on whether the result set contains an unnamed attack. REST types the same field as optional and is unaffected. This is exactly the class of thing a single source of truth for output shapes exists to prevent — the schema said one thing, the data another, and a consumer found out at runtime.

### A case study: `localId`

The reference docs type `localId` as "String or Number". I'd typed it `string` — every card the live API serves quotes it (I checked the full card list, zero unquoted occurrences), and the GraphQL schema says `String!`. But in C# that mismatch isn't a narrowing inconvenience, it's a thrown exception, and because the property is required, one unquoted value would fail deserialization of the whole card. I shipped 0.1.1 to accept both. A published schema is the difference between finding that by re-reading docs (which is how I found it) and finding it in CI.

### On open vs. closed enums, from a strict-language consumer

Your call to keep `rarity`/`stage`/`suffix` open is right, and I can offer proof it's not hypothetical: TCG Pocket arrived carrying an entirely new rarity vocabulary (`One Diamond` … `Crown`). My SDK deserializes all of it only because rarity is a string. Any client that had closed that enum would have thrown on every Pocket card.

One caution on the closed set (`category`, `types`, `trainerType`, …): in C#, a real `enum` **throws** on an unknown value — a harder failure than a TS union that merely fails to narrow. So a closed enum in the schema becomes brittle in strict languages even when the value set is genuinely stable, if a generator maps it naively. Two cheap mitigations: carry the known values as `x-extensible-enum`-style metadata even on closed enums so generators for strict languages can choose a string-backed constants type, or note the recommendation in the generator guidance. (Related: `weaknesses[].value` is one string field whose semantics fork by game — `×2` physical, `+20` Pocket — which is the kind of thing schema descriptions could carry even where types can't.)

### An offer

When Stage 1 publishes `openapi.json` as a release artifact, I'd like to wire a CI job that **validates my hand-written models against it** rather than generating from it (generation would flatten my source-generated serialization and AOT guarantees). I think that's also a better test of the schema than a generated client: an independent implementation agreeing with the contract proves more than the contract round-tripping its own output. Happy to be the guinea pig for the C#/.NET row of your language table.

PR #1722, opened 2026-07-13
feat: update sv01 card variants to detailed variant format

Octopixell:
## Changes

Updated all variant data for SV01 in the new detailed variant format

---

P.S. Sorry for the two earlier PR's with single card changes- It was after those two PR's I found out the whole set needed checking. 😅


data/Scarlet & Violet/Scarlet & Violet/001.ts
```diff
@@ -53,9 +53,16 @@ const card: Card = {
 	retreat: 2,
 	regulationMark: "G",
 
-	variants: {
-		holo: false
-	},
+	variants: [
+		{
+			type: "normal",
+			size: "standard"
```
FalconChipp: Standard is the default size so there's no need for defining it for each. 

Needed only when the size changes to jumbo

FalconChipp (changes requested):

data/Scarlet & Violet/Scarlet & Violet/001.ts
Octopixell: I'll change that 👍

Octopixell: Fixed 👍

FalconChipp (approved):

PR #1742, opened 2026-07-16
(data) Add Japanese SM set SM10a GG End

KlausDerKleber:
## What

Adds the complete Japanese set **SM10a ジージーエンド (GG End)**, released 2019-04-05:

- `data-asia/SM/SM10a/001.ts` … `069.ts` — all 69 cards (54 regular + 15 secret rares: 8 SR, 4 UR, 3 Secret Rare)
- the set definition `data-asia/SM/SM10a.ts` already existed and is untouched

## Data sources

- **pokemon-card.com** (official database): Japanese names, flavor texts, National Dex numbers, official card count
- **limitlesstcg.com**: full Japanese card text, stages, weaknesses/resistances/retreat, rarities, illustrators, attack costs
- **Cardmarket**: product IDs as `thirdParty.cardmarket` on the variant objects (557278–557348; tcgplayer IDs not available to me)

## Notes

- Follows the file format and conventions of the M-era sets (#1728)
- All 69 files type-check against `interfaces.d.ts` (`tsc --noEmit`)


KlausDerKleber:
CI was failing because the compiler resolves **both** the set definition file (`getSet`) and the card folder (`getCard`) from `set.id`, and `data-asia/SM/SM10a.ts` carried `id: 'sn10a'`. As soon as the card folder existed the compile broke with:

```
error: Cannot find module '../../..//data-asia/SM/sn10a/047.ts' from 'server/compiler/utils/cardUtil.ts'
```

Fixed by correcting the id to `SM10a`, which matches the file name, the card folder and the neighbouring sets (SM10, SM10b, SM11a, SM11b). This does change the public set id `sn10a` -> `SM10a`; that set currently exposes 0 cards and has no assets keyed under `sn10a`, so nothing else depends on it. Renaming the card folder to `sn10a` instead is not an option — `getSet` would then look for `data-asia/SM/sn10a.ts`, which does not exist.

Verified locally: `bun run compile`, root `bun run validate` and server `bun run --bun validate` all pass; `generated/ja/sets.json` now lists SM10a with 69 cards.

PR #1744, opened 2026-07-16
(data) Add Japanese SM set SM11 Miracle Twin

KlausDerKleber:
## What

Adds the complete Japanese set **SM11 ミラクルツイン (Miracle Twin)**, released 2019-05-31:

- `data-asia/SM/SM11/001.ts` … `115.ts` — all 115 cards (94 regular + 21 secret rares: 12 SR, 3 Secret Rare, 6 UR)
- the set definition `data-asia/SM/SM11.ts` already existed and is untouched

## Data sources

- **pokemon-card.com** (official database): Japanese names, flavor texts, National Dex numbers, official card count
- **limitlesstcg.com**: full Japanese card text, stages, weaknesses/resistances/retreat, rarities, illustrators, attack costs
- **Cardmarket**: product IDs as `thirdParty.cardmarket` on the variant objects (556960–557077; tcgplayer IDs not available to me)

## Notes

- Follows the file format and conventions of the M-era sets (#1728)
- All 115 files type-check against `interfaces.d.ts` (`tsc --noEmit`)


KlausDerKleber:
CI was failing because the compiler resolves **both** the set definition file (`getSet`) and the card folder (`getCard`) from `set.id`, and `data-asia/SM/SM11.ts` carried `id: 'sn11'`. As soon as the card folder existed the compile broke with:

```
error: Cannot find module '../../..//data-asia/SM/sn11/047.ts' from 'server/compiler/utils/cardUtil.ts'
```

Fixed by correcting the id to `SM11`, which matches the file name, the card folder and the neighbouring sets (SM10, SM10b, SM11a, SM11b). This does change the public set id `sn11` -> `SM11`; that set currently exposes 0 cards and has no assets keyed under `sn11`, so nothing else depends on it. Renaming the card folder to `sn11` instead is not an option — `getSet` would then look for `data-asia/SM/sn11.ts`, which does not exist.

Verified locally: `bun run compile`, root `bun run validate` and server `bun run --bun validate` all pass; `generated/ja/sets.json` now lists SM11 with 115 cards.

PR #1813, opened 2026-07-23
(data) en Call of Legends col1 Added Missing Variants 

advra:
closes #N/A

## Changes

- add missing variants for entire set 
- added tcgplayerids

## Details

- added new subtype `phanphy error` for 66 (both normal and reverse)
image
image


thomas-bassett:
please fix conflicts

advra:
ok i added more variants. should be good to go when tests pass

PR #1826, opened 2026-07-24
(data) en tk-xy-latia Trainer Kit Ex XY Latia Add Missing Data

advra:
closes #N/A

## Changes

- initially 30 cards
- corrected to 10 cards (via reference https://bulbapedia.bulbagarden.net/wiki/EX_Trainer_Kit_(TCG)) 

## Details

- removed cards 11-30 (wrong?)
- added tcgplayer variants and missing data


advra:
wrong changes

PR #1836, opened 2026-07-25
(data) en ex-tk-latia EX Trainer Kit Latias - added card variants 

advra:
closes #N/A

## Changes

- moved thirdparty to variants
- moved tcgplayer ids
- added card market ids

## Details

- fixed incorrect weakness from x2 to just x1

### DIFF
--- a/data/Trainer kits/EX trainer Kit (Latias)/10.ts
+++ b/data/Trainer kits/EX trainer Kit (Latias)/10.ts
@@ -21,6 +21,7 @@ const card: Card = {
 		{
 			type: "normal",
 			thirdParty: {
+				cardmarket: 275771,
 				tcgplayer: 85457
 			}
 		},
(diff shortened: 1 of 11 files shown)


data/Trainer kits/EX trainer Kit (Latias)/10.ts
```diff
@@ -7,16 +7,25 @@ const card: Card = {
...
-	}
+	types: [
+		"Fire"
+	],
+	retreat: 0,
+	variants: [
+		{
+			type: "normal",
+			thirdParty: {
+				tcgplayer: 85457
```
FalconChipp, line 25: Missing cardmarket ID 

275771
https://www.cardmarket.com/en/Pokemon/Products/Singles/EX-Trainer-Kit/Fire-Energy-TK1Latias-10

FalconChipp (changes requested):

data/Trainer kits/EX trainer Kit (Latias)/10.ts
advra: very nice catch. thanks!

FalconChipp (approved):

PR #1840, opened 2026-07-25
(data) en ex1 Ruby & Sapphire - added card variants 

advra:
closes #N/A
## Changes

- added card market ids
- added tcgplayer ids

## Details

- added missing jumbos
- added missing tournaments
- added missing winner variants

### DIFF
--- a/data/EX/Ruby & Sapphire/1.ts
+++ b/data/EX/Ruby & Sapphire/1.ts
@@ -9,18 +9,16 @@ const card: Card = {
 	},
 
 	illustrator: "Mitsuhiro Arita",
-	rarity: "Rare",
+	rarity: "Holo Rare",
 	category: "Pokemon",
 	set: Set,
 
-	dexId: [
-		306,
-	],
+	dexId: [306],
 
 	hp: 110,
 
 	types: [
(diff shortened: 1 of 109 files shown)


data/EX/Ruby & Sapphire/1.ts
```diff
@@ -103,19 +101,23 @@ const card: Card = {
...
+			thirdParty: {
+				cardmarket: 275649,
+				tcgplayer: 83475
+			}
 		},
 		{
 			type: "reverse",
+			thirdParty: {
+				tcgplayer: 83475
+			}
```
FalconChipp, line 118: Cardmarket uses the same ID for reverses too, so please update with the cardmarket IDs set wide

FalconChipp (changes requested):

data/EX/Ruby & Sapphire/1.ts
advra: good catch. tcgplayer does the same. didnt realize cardmarket was like that. makes things easier! 

ive added all reverses now

FalconChipp (approved):

PR #1844, opened 2026-07-26
(data) en tk-hs-r Trainer Kit Raichu - added card variants 

advra:
closes #N/A

## Changes

- added missing data from cards 1-30
- add tcgplayer ids
- add cardmarket ids
- moved thirdparty
-

### DIFF
--- a/data/Trainer kits/HS trainer Kit (Raichu)/19.ts
+++ b/data/Trainer kits/HS trainer Kit (Raichu)/19.ts
@@ -4,6 +4,7 @@ import Set from '../HS trainer Kit (Raichu)'
 const card: Card = {
 	name: {
 		en: "Raichu",
+		fr: "Raichu"
 	},
 
 	illustrator: "match",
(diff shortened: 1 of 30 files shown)


data/Trainer kits/HS trainer Kit (Raichu)/19.ts
```diff
@@ -49,9 +49,6 @@ const card: Card = {
 		}
 	}],
 
-	thirdParty: {
-		tcgplayer: 88520
-	}
```
FalconChipp, line 54: Variants block missing

FalconChipp (changes requested):

data/Trainer kits/HS trainer Kit (Raichu)/19.ts
advra: added the thirdparty variants

FalconChipp (approved):

PR #1847, opened 2026-07-26
(data) en hgss Heart Gold Soul Silver - added card variants 

advra:
closes #N/A

## Changes

- add card market ids
- add tcpglayer ids

### DIFF
--- a/data/HeartGold & SoulSilver/HeartGold SoulSilver/100.ts
+++ b/data/HeartGold & SoulSilver/HeartGold SoulSilver/100.ts
@@ -25,6 +25,7 @@ const card: Card = {
 		{
 			type: "normal",
 			thirdParty: {
+				cardmarket: 279072,
 				tcgplayer: 88392
 			}
 		},
@@ -49,6 +50,7 @@ const card: Card = {
 			}
 		}
 	],
+
 	hp: 0
 	retreat: 0
 }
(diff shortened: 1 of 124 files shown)


data/HeartGold & SoulSilver/HeartGold SoulSilver/100.ts
```diff
@@ -15,30 +15,43 @@ const card: Card = {
...
 
+	hp: 0,
+	retreat: 0,
 	variants: [
 		{
-			type: "normal"
+			type: "normal",
+			thirdParty: {
+				tcgplayer: 88392
+			}
```
FalconChipp, line 32: Cardmarket IDs missing 
https://www.cardmarket.com/en/Pokemon/Products/Singles/HeartGold-SoulSilver?idRarity=0&sortBy=collectorsnumber_asc&perSite=30

FalconChipp (changes requested):
I have added a link to the HGSS set on cardmarket as there seems to be a fair few cards missing cardmarket IDs like the sigs in 101 & 103 for example as well as normals and so on in other cards

data/HeartGold & SoulSilver/HeartGold SoulSilver/100.ts
advra: Added all the missing cardmarket variants

FalconChipp (approved):

PR #1848, opened 2026-07-26
(data) en hgss3 Undaunted - added card variants 

advra:
closes #N/A

## Changes

- add cardmarket
- add tcgplayer variants
- moved thirdparty into per variants


advra:
I noticed some formatting on a few files are bad. will need to fix those

advra:
ok should be good. i went through and added all the variants and formatted

PR #1849, opened 2026-07-26
(data) ja SV1V Violet ex - cardmarket + tcgplayer product ids

KlausDerKleber:
## What

Links the cards of the existing Japanese set **SV1V バイオレットex (Violet ex)** to their Cardmarket **and TCGplayer** products, so the API can serve their prices.

- **68** cards carried no product id at all and get both
- **40** carried a card-level `thirdParty.cardmarket`, which is moved onto the card's `variants` entry and joined by its tcgplayer id
- **7** carried the cardmarket id of a different print and are corrected

Afterwards every card of the set carries its id on a `variants` entry and none on the card itself, which is where tcgdex is heading. Nothing else changes: the `zh-tw`/`th`/`id` texts and the Japanese card data are not rewritten.

7 cards carried the product of **another print** of the same card, so the API served that print's price for them:

| card | was | is | the old id belongs to |
|---|---|---|---|
| 082 | 693093 | 693157 | points at card 019 of this set |
| 083 | 693095 | 693158 | points at card 021 of this set |
| 084 | 693104 | 693159 | points at card 029 of this set |
| 086 | 693113 | 693161 | points at card 038 of this set |
| 092 | 693090 | 693167 | points at card 016 of this set |
| 093 | 693103 | 693168 | points at card 028 of this set |
| 095 | 693116 | 693170 | points at card 041 of this set |

The ids are assigned **by collection number**, not by name: within a Cardmarket expansion the products sorted by `idProduct` follow the collection order, and the assignment is verified against the sample rows pokepricelab renders. That matters because Cardmarket names every print of a card identically inside an expansion, so a name lookup cannot tell a base print from its AR/full-art reprint — which is where the corrected ids above come from. That the 33 ids this set already carried correctly agree with the same assignment is an independent cross-check.

## Data sources

- **Cardmarket** public product catalog (`downloads.s3.cardmarket.com`): the product ids (693075–693183), assigned by collection number
- **TCGplayer** via [tcgcsv.com](https://tcgcsv.com) (category 85, "Pokemon Japan"): product IDs as `thirdParty.tcgplayer` (568227–568334), keyed by the collection number each product carries in `extendedData`

## Notes

- `tsc --noEmit` reports the same 30 pre-existing errors as upstream/master — this change adds none. They are `rarity` being required by `interfaces.d.ts` while these older hand-written files omit it, which is out of scope here.

### DIFF
--- a/data-asia/SV/SV1V/001.ts
+++ b/data-asia/SV/SV1V/001.ts
@@ -52,12 +52,18 @@ const card: Card = {
 		value: "×2"
 	}],
 
+	variants: [
+		{
+			type: "normal",
+			thirdParty: {
+				cardmarket: 693075,
+				tcgplayer: 568227,
+			},
+		},
+	],
+
 	retreat: 2,
 	regulationMark: "G",
-
-	thirdParty: {
-		cardmarket: 693075
-	}
 }
 
 export default card
--- a/data-asia/SV/SV1V/002.ts
+++ b/data-asia/SV/SV1V/002.ts
@@ -63,12 +63,18 @@ const card: Card = {
 		value: "×2"
 	}],
 
+	variants: [
+		{
+			type: "normal",
+			thirdParty: {
+				cardmarket: 693076,
+				tcgplayer: 568228,
+			},
+		},
+	],
+
 	retreat: 2,
 	regulationMark: "G",
-
-	thirdParty: {
-		cardmarket: 693076
-	}
 }
 
 export default card
--- a/data-asia/SV/SV1V/003.ts
+++ b/data-asia/SV/SV1V/003.ts
@@ -63,12 +63,18 @@ const card: Card = {
 		value: "×2"
 	}],
 
+	variants: [
+		{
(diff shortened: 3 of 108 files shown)


data-asia/SV/SV1V/001.ts
```diff
@@ -52,12 +52,17 @@ const card: Card = {
 		value: "×2"
 	}],
 
+	variants: [
+		{
+			type: "normal",
+			thirdParty: {
+				cardmarket: 693075,
```
advra, line 59: Please also add tcgplayer ids for variants https://www.tcgplayer.com/search/pokemon-japan/sv1v-violet-ex?productLineName=pokemon-japan&productTypeName=Cards&page=1&view=grid&setName=sv1v-violet-ex

KlausDerKleber: They are not needed per se - there are multiple sets without them. I've added them nevertheless to also cater for the US ppl.

PR #1850, opened 2026-07-26
(data) ja M1S Mega Symphonia - cardmarket + tcgplayer product ids

KlausDerKleber:
## What

Links the cards of the existing Japanese set **M1S メガシンフォニア (Mega Symphonia)** to their Cardmarket **and TCGplayer** products, so the API can serve their prices.

- **92** cards carried no product id at all and get both

Afterwards every card of the set carries its id on a `variants` entry and none on the card itself, which is where tcgdex is heading. Nothing else changes.

The ids are assigned **by collection number**, not by name: within a Cardmarket expansion the products sorted by `idProduct` follow the collection order, and the assignment is verified against the sample rows pokepricelab renders. That matters because Cardmarket names every print of a card identically inside an expansion, so a name lookup cannot tell a base print from its AR/full-art reprint. This set carried no id yet, so there is nothing in the repository to cross-check the assignment against — it rests on the verified ordering alone.

## Data sources

- **Cardmarket** public product catalog (`downloads.s3.cardmarket.com`): the product ids (840539–840651), assigned by collection number
- **TCGplayer** via [tcgcsv.com](https://tcgcsv.com) (category 85, "Pokemon Japan"): product IDs as `thirdParty.tcgplayer` (647202–647293), keyed by the collection number each product carries in `extendedData`

## Notes

- All 92 files type-check against `interfaces.d.ts` (`tsc --noEmit`)

### DIFF
--- a/data-asia/M/M1S/001.ts
+++ b/data-asia/M/M1S/001.ts
@@ -33,6 +33,16 @@ const card: Card = {
 		type: "Fire",
 		value: "×2"
 	}],
+	variants: [
+		{
+			type: "normal",
+			thirdParty: {
+				cardmarket: 840539,
+				tcgplayer: 647258,
+			},
+		},
+	],
+
 	retreat: 2,
 	regulationMark: "I",
 	rarity: "Common",
--- a/data-asia/M/M1S/002.ts
+++ b/data-asia/M/M1S/002.ts
@@ -14,6 +14,16 @@ const card: Card = {
 		ja: "暖かい土地で暮らすものほど、ツルの伸びが早い。伸びると自ら切って短くする。"
 	},
 	stage: "Stage1",
+	variants: [
+		{
+			type: "normal",
+			thirdParty: {
+				cardmarket: 840540,
+				tcgplayer: 647259,
+			},
+		},
+	],
+
 	evolveFrom: {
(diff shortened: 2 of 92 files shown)


advra:
Please also add tcgplayerids https://www.tcgplayer.com/search/pokemon-japan/m1s-mega-symphonia?productLineName=pokemon-japan&productTypeName=Cards&page=1&view=grid&setName=m1s-mega-symphonia

ceesvanegmond:
Is it true that this is not on production yet?

For instance https://api.tcgdex.net/v2/en/cards/m1s-001 doesn't seem to have cardmarket nor tcgplayer values 

@KlausDerKleber

KlausDerKleber:
> Is it true that this is not on production yet?
> 
> For instance https://api.tcgdex.net/v2/en/cards/m1s-001 doesn't seem to have cardmarket nor tcgplayer values
> 
> @KlausDerKleber

No, there was no new release of tcgdex since merge. 
Not sure when there will be one, @thomas-bassett should know 😄 

(also you may want to use /ja/ instead of /en/) 😬 )

ceesvanegmond:
@KlausDerKleber touche....

@thomas-bassett When can we expect a new release?

PR #1852, opened 2026-07-26
(data) ja SV9 Battle Partners - cardmarket + tcgplayer product ids

KlausDerKleber:
## What

Links the cards of the existing Japanese set **SV9 バトルパートナーズ (Battle Partners)** to their Cardmarket **and TCGplayer** products, so the API can serve their prices.

- **69** cards carried no product id at all and get both
- **63** carried a card-level `thirdParty.cardmarket`, which is moved onto the card's `variants` entry and joined by its tcgplayer id
- **12** carried the cardmarket id of a different print and are corrected

Afterwards every card of the set carries its id on a `variants` entry and none on the card itself, which is where tcgdex is heading. Nothing else changes: the `zh-tw`/`zh-cn` texts and the Japanese card data are not rewritten.

12 cards carried the product of **another print** of the same card, so the API served that print's price for them:

| card | was | is | the old id belongs to |
|---|---|---|---|
| 101 | 807593 | 807688 | points at card 006 of this set |
| 102 | 807605 | 807689 | points at card 018 of this set |
| 103 | 807612 | 807690 | points at card 025 of this set |
| 106 | 807631 | 807693 | points at card 044 of this set |
| 107 | 807639 | 807694 | points at card 052 of this set |
| 110 | 807664 | 807697 | points at card 077 of this set |
| 111 | 807667 | 807698 | points at card 080 of this set |
| 113 | 807604 | 807700 | points at card 017 of this set |
| 116 | 807633 | 807703 | points at card 046 of this set |
| 119 | 807659 | 807706 | points at card 072 of this set |
| 124 | 807604 | 807711 | points at card 017 of this set |
| 129 | 807659 | 807716 | points at card 072 of this set |

The ids are assigned **by collection number**, not by name: within a Cardmarket expansion the products sorted by `idProduct` follow the collection order, and the assignment is verified against the sample rows pokepricelab renders. That matters because Cardmarket names every print of a card identically inside an expansion, so a name lookup cannot tell a base print from its AR/full-art reprint — which is where the corrected ids above come from. That the 51 ids this set already carried correctly agree with the same assignment is an independent cross-check.

## Data sources

- **Cardmarket** public product catalog (`downloads.s3.cardmarket.com`): the product ids (807588–807719), assigned by collection number
- **TCGplayer** via [tcgcsv.com](https://tcgcsv.com) (category 85, "Pokemon Japan"): product IDs as `thirdParty.tcgplayer` (614852–615008), keyed by the collection number each product carries in `extendedData`

## Notes

- All 132 files type-check against `interfaces.d.ts` (`tsc --noEmit`)

### DIFF
--- a/data-asia/SV/SV9/001.ts
+++ b/data-asia/SV/SV9/001.ts
@@ -42,12 +42,18 @@ const card: Card = {
 		value: "×2"
 	}],
 
+	variants: [
+		{
+			type: "normal",
+			thirdParty: {
+				cardmarket: 807588,
+				tcgplayer: 614852,
+			},
+		},
+	],
+
 	retreat: 1,
 	regulationMark: "I",
-
-	thirdParty: {
-		cardmarket: 807588
-	}
 }
 
 export default card
--- a/data-asia/SV/SV9/002.ts
+++ b/data-asia/SV/SV9/002.ts
@@ -46,12 +46,18 @@ const card: Card = {
 		value: "×2"
 	}],
 
+	variants: [
+		{
+			type: "normal",
+			thirdParty: {
+				cardmarket: 807589,
+				tcgplayer: 614853,
+			},
+		},
+	],
+
 	retreat: 3,
 	regulationMark: "I",
-
-	thirdParty: {
-		cardmarket: 807589
-	}
 }
 
 export default card
--- a/data-asia/SV/SV9/003.ts
+++ b/data-asia/SV/SV9/003.ts
@@ -48,12 +48,18 @@ const card: Card = {
 		value: "×2"
 	}],
 
+	variants: [
+		{
+			type: "normal",
+			thirdParty: {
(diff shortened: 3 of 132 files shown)


advra:
Please add tcgplayer ids for all the variants https://www.tcgplayer.com/search/pokemon-japan/sv9-battle-partners?productLineName=pokemon-japan&setName=sv9-battle-partners&productTypeName=Cards&page=1&view=grid

data-asia/SV/SV9/001.ts
```diff
@@ -42,12 +42,17 @@ const card: Card = {
 		value: "×2"
 	}],
 
+	variants: [
+		{
+			type: "normal",
+			thirdParty: {
+				cardmarket: 807588,
```
advra, line 49: cards all need tcgplayer ids

PR #1853, opened 2026-07-26
(data) ja SV8a Terastal Fest ex - cardmarket + tcgplayer product ids

KlausDerKleber:
## What

Links the cards of the existing Japanese set **SV8a テラスタルフェスex (Terastal Fest ex)** to their Cardmarket **and TCGplayer** products, so the API can serve their prices.

- **237** cards carried no product id at all and get both

Afterwards every card of the set carries its id on a `variants` entry and none on the card itself, which is where tcgdex is heading. Nothing else changes: the `zh-tw`/`zh-cn`/`th`/`id` texts and the Japanese card data are not rewritten.

The ids are assigned **by collection number**, not by name: within a Cardmarket expansion the products sorted by `idProduct` follow the collection order, and the assignment is verified against the sample rows pokepricelab renders. That matters because Cardmarket names every print of a card identically inside an expansion, so a name lookup cannot tell a base print from its AR/full-art reprint. This set carried no id yet, so there is nothing in the repository to cross-check the assignment against — it rests on the verified ordering alone.

## Data sources

- **Cardmarket** public product catalog (`downloads.s3.cardmarket.com`): the product ids (803114–803350), assigned by collection number
- **TCGplayer** via [tcgcsv.com](https://tcgcsv.com) (category 85, "Pokemon Japan"): product IDs as `thirdParty.tcgplayer` (602319–602701), keyed by the collection number each product carries in `extendedData`

## Notes

- All 237 files type-check against `interfaces.d.ts` (`tsc --noEmit`)

### DIFF
--- a/data-asia/SV/SV8a/001.ts
+++ b/data-asia/SV/SV8a/001.ts
@@ -54,6 +54,29 @@ const card: Card = {
 		value: "×2"
 	}],
 
+	variants: [
+		{
+			type: "normal",
+			thirdParty: {
+				cardmarket: 803114,
+				tcgplayer: 602319,
+			},
+		},
+		{
+			type: "reverse",
+			foil: "masterball",
+			thirdParty: {
+				tcgplayer: 604502,
+			},
+		},
+		{
+			type: "reverse",
+			thirdParty: {
+				tcgplayer: 602320,
+			},
+		},
+	],
+
 	retreat: 0,
 	regulationMark: "H"
 }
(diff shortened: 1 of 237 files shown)


advra:
Please also add tcgplayer ids

data-asia/SV/SV8a/001.ts
```diff
@@ -54,6 +54,15 @@ const card: Card = {
 		value: "×2"
 	}],
 
+	variants: [
+		{
+			type: "normal",
+			thirdParty: {
+				cardmarket: 803114,
```
advra, line 61: This needs tcgplayerids https://www.tcgplayer.com/search/pokemon-japan/sv8a-terastal-fest-ex?productLineName=pokemon-japan&productTypeName=Cards&page=1&view=grid&setName=sv8a-terastal-fest-ex

PR #1891, opened 2026-07-27
(data) en ex5 Ex Hidden Legends - Added Missing Variants + CardMarket + Tcgplayer Ids 

advra:
closes #N/A

## Changes

- added cardmarket and tcgplayerid

## Details

- referenced variants
- added championships stamps and sigs

### DIFF
--- a/data/EX/Hidden Legends/1.ts
+++ b/data/EX/Hidden Legends/1.ts
@@ -92,6 +92,7 @@ const card: Card = {
 		},
 		{
 			type: "reverse",
+			foil: 'energy',
 			thirdParty: {
 				tcgplayer: 83719,
 				cardmarket: 276075
--- a/data/EX/Hidden Legends/1.ts
+++ b/data/EX/Hidden Legends/1.ts
@@ -97,8 +97,7 @@ const card: Card = {
 				cardmarket: 276075
 			}
 		},
-	],
-
+	]
 }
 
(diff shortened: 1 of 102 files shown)


advra:
EX era are a bit quirky with reverse. I change Holo Energy to be Reverse Energy. ie Cards 1-27 that are reverse holo variant have energy symbols behind text. For Example the card 1 bennet holo. However when observing cards 28 to 92 I noticed we initially had holo which is incorrect. The correct typing for EX era is reverse holos despite not have the middle energy behind text and having the usual "holo" pattern we typically see in other eras. Example from 92 below

|id| Card Name | Notes| Image|
|--|----------|---------------|-------|
|1| Benette (Holo) | Holo card variant shows no energy | image|
|1| Benette (Reverse Holo) | Card variant is is categorized as reverse holo (from pkmn.gg pricecharting and ebay listings) featuring an energy symbol behind description text | image |
|92| Steven's Advice (Normal) | Card variant is a typical normal card | image |
|92| Steven's Advice (Reverse Holo) | Card variant is is categorized as reverse holo (from pkmn.gg pricecharting and ebay listings) rather than holo. Other collecting apps such as Collectr show 92 has Normal and Reverse variants. This would be correctly typed as Holo if this was a non-EX era set (D&P/Platinum and newer).  | image |

PR #1914, opened 2026-08-04
Add missing German translations for base1

Que-tin:
### How and why?

I was playing around with the databse, when I noticed I can't find plenty of my
german cards due to lacking docs or because there was english text in the german
card docs.

So I build a tool locally (with AI) to check for german gaps and fill them up
with actual authentic card data. To make sure this data is accurate I'm using
PokéWiki (large german card database) + OCR on card screenshots + the
`NSSpellChecker` with the German dictionary to make sure no mistakes from the
wiki or the scan get into the files.

I will create plenty of PRs because the german documentation right now is far
from complete. If you have any question let me know. Where changes come from
etc. is fully logged on my device and I can tell you where which text comes
from.


data/Base/Base Set/10.ts
```diff
@@ -61,7 +61,7 @@ const card: Card = {
 			effect: {
 				en: "Discard 1 Energy card attached to Mewtwo in order to use this attack. During your opponent's next turn, prevent all effects of attacks, including damage, done to Mewtwo.",
 				fr: "Défaussez 1 carte Énergie  attachée à Mewtwo pour pouvoir utiliser cette attaque. Pendant le prochain tour de votre adversaire, prévenez tous les effets ou attaques, y compris les dégâts, infligés à Mewtwo.",
-				de: "Entferne eine auf Mewtu abgelegte  Energiekarte um diesen Angriff auszuführen. Verhindere während desnächsten gegnerischen Zuges alle Auswirkungen von Angriffen auf Mewtu (einschließlich der Schadenspunkte)",
+				de: "Entferne eine auf Mewtu abgelegte Psycho Energiekarte, um diesen Angriff auszuführen. Verhindere während des nächsten gegnerischen Zugs alle Auswirkungen von Angriffen auf Mewtu (einschließlich der Schadenspunkte).",
```
thomas-bassett: `Psycho` is that a symbol or text. i think if your putting in the effort to get this right this should use something like `{P}` i believe is the code used for Psychic energy symbols over text.

thomas-bassett (commented):
please check all the energies, best todo this once now. this is likely the same in the other prs

thomas-bassett (dismissed):
please check all the energies, best todo this once now. this is likely the same in the other prs

FalconChipp (approved):

PR #1923, opened 2026-08-06
Remove incorrect normal variants from ME05

Octopixell:
- Remove normal variant from me05-028 (Miraidon)
- Remove normal variant from me05-035 (Spiritomb)
- Remove normal variant from me05-056 (Zarude)

### DIFF
--- a/data/Mega Evolution/Pitch Black/028.ts
+++ b/data/Mega Evolution/Pitch Black/028.ts
@@ -82,13 +82,6 @@ const card: Card = {
 
 	variants: [
 		{
-			type: "normal",
-			thirdParty: {
-				cardmarket: 895815,
-				tcgplayer: 704785
-			}
-		},
-		{
 			type: "reverse",
 			thirdParty: {
 				cardmarket: 895815,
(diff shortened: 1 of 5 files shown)


data/Mega Evolution/Pitch Black/028.ts
```diff
@@ -81,13 +81,6 @@ const card: Card = {
 	regulationMark: "J",
 
 	variants: [
-		{
-			type: "normal",
```
thomas-bassett, line 85: This was included in a build and battleset so is a legit card

data/Mega Evolution/Pitch Black/035.ts
```diff
@@ -59,13 +59,6 @@ const card: Card = {
 	regulationMark: "J",
 
 	variants: [
-		{
-			type: "normal",
```
thomas-bassett, line 63: This was included in a build and battleset so is a legit card

thomas-bassett (changes requested):

data/Mega Evolution/Pitch Black/028.ts
Octopixell: Eeehm, shouldn't it then be part of that set? And not this set? There's no way to differentiate that right now?

PR #1925, opened 2026-08-06
feat: Add VariantLine type to card variant_detailed

thomas-bassett:
## Changes
Adds `line` to variant detailed. this line is used to separate variants based on how they are obtained. raised out of the issue of trying to distinguish between a holo released in booster packs and a non-holo variant released in build and battle kits. 

This defines four line types:
**Main**: found in booster packs
**Parallel**: found in specific products bbk, pre constructed
**Restricted**: given out at a specific event or promotion
**Exclusive**: given out at a specific event for either working, topcut, winners. or for performing a specific job: professor program.

variants currently default to main line at compile time. so in theory only need to adjust the line for the other 3

none of these lines are locked in a new reasons can be added to expand on them. if a card is available in two of the lines always use the more available line. i.e. precons will likely have main line cards in them. those cards should be treated as main line and the non-holo holos should be treated as parallel as they are not available in booster packs

### DIFF
--- a/interfaces.d.ts
+++ b/interfaces.d.ts
@@ -16,6 +16,7 @@ export interface Serie {
 	energies?: Array<Types>
 }
 
+export type VariantLine = 'main' | 'parallel' | 'restricted' | 'exclusive'
 export type VariantType =  'normal' | 'holo' | 'reverse' | 'metal' | 'lenticular'
 export type VariantStamps = '1st-edition' | 'w-promo' | 'pre-release' | 'pokemon-center' | 'set-logo' | 'staff' | 'pikachu-tail'
 	| 'wotc' | 'd-edition-error' | '1st-edition-scratch-error' | "1st-edition-error" | '1st-movie' | '1st-movie-inverted'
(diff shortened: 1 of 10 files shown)


cloetensbrecht:
I like the general idea of introducing `line`. I think the distinction between `main`, `parallel`, `restricted` and `exclusive` is useful, but I'm wondering if `line` should contain a bit more detail.

At the moment, the `line` value tells us the general category a variant belongs to, but not the specific distribution. For example, a Deck Exclusive and a Prize Pack card could both belong to a non-main line, but from `line` alone we cannot determine which specific distribution applies.

Would it make sense to make `line` an array containing structured objects, for example:
```ts
line: [{
  category: 'restricted',
  distribution: ['prize-pack']
}]
```
or
```ts
line: [{
  category: 'parallel',
  distribution: ['deck-exclusive']
}]
```

This would make `category` the high-level classification, while `distribution` provides the specific distribution method.

This also keeps the categories relatively broad instead of requiring a new category for every specific distribution type. It would allow us to distinguish regular/main-line variants from specially distributed variants while still retaining the specific distribution information.

I think this could also make it easier to support additional distribution types in the future without making the `category` values increasingly specific.

Would this be a reasonable direction for the `line` structure? I think it would give us the general distinction provided by `line`, while still retaining the more detailed information about how a variant was distributed.

thomas-bassett:
So with the distribution it's some good data and we should have it but I think we can do better than a tag. 

I think this links into when we get sealed/physical. We would be able to directly link the variant to the exact prize-pack that is comes out from. I belive that would cover the distribution your after.

Is there somthing I've missed about your solution?

cloetensbrecht:
I can follow you on that. I agree that distribution shouldn't just be a simple tag. Linking the variant directly to the relevant sealed/physical product would actually be a much better and more complete solution.

One thing I think this also raises is the existing `boosters` property on the Card itself. I think that should probably eventually move down to the variant level as well.

The reason is that not every variant of a card comes from the same boosters/products. For example, with Eldegoss 16/203, the regular print, the Deck Exclusive print and the Prize Pack print have different sources. With boosters currently being defined on the card level, we can't distinguish which variant actually comes from which booster/product.

So I think the current logic is actually incorrect for cards with multiple distribution variants: all variants effectively inherit the same boosters, even though some of them were never available in those boosters.

thomas-bassett:
I believe the boosters property was specific to `pocket` cards, was added before my time so im not fully aware.

but when we get to adding sealed it will be something we come back to revisit, the way i picture it now is that the `variant` would have a `links` field that would contain some more specific data. would be laid out something like this: 

```
"links": {
        "type": "distribution",
        "name": "prize-pack-3",
        "path": "/v3/sealed/misc/pp3"
    }
```

I just threw this together so don't treat it as gospel

thomas-bassett:
This is going to be address in `V3` closing for now.

PR #1926, opened 2026-08-06
(fix) Feraligatr - English was marked as german and english was missing

Que-tin:


thomas-bassett (changes requested):
I know your planning to fix the de translations but is there a reason we are removing the de here? rather than just updating it?

thomas-bassett (approved):

PR #1966, opened 2026-08-12
(fix) Dark Explorers: typo in the German set name

Que-tin:
The German name of `bw5` is missing an `r`:

```
de: "Erfoscher der Finsternis",
```

It should be **Erforscher** der Finsternis — from *erforschen*, to explore.


Que-tin:
@FalconChipp

FalconChipp (approved):

FalconChipp (approved):

PR #1972, opened 2026-08-12
(fix) Remove French-only duplicate attack entries from 45 cards

Que-tin:
Forty-five cards carry an extra `attacks` entry that duplicates an attack already on the card, with every language but French stripped out. `data/Sun & Moon/Cosmic Eclipse/210.ts` (Rillaboom & Cinderace & Inteleon-GX):

```ts
{
    cost: ["Colorless", "Colorless", "Colorless"],
    name:   { en: "Solar Plant GX", fr: "Plante Solaire GX", …, de: "Solargewächs GX" },
    effect: { en: "This attack does 50 damage to each of your opponent’s Pokémon. …", … }
},
{
    cost: ["Colorless", "Colorless", "Colorless"],   // ← the same attack again
    name:   { fr: "Plante Solaire-GX" },
    effect: { fr: "Cette attaque inflige 50 dégâts à chacun des Pokémon de votre adversaire. …" }
}
```

Same cost, same damage, same French effect text — and on the GX cards the stub writes `Name-GX` where the real entry writes `Name GX`.

Every stub was required to agree with its twin on **cost and damage**, and on **either** the French name ignoring punctuation **or** the French effect text verbatim, before removal. None is a distinct attack that merely lacks English.

The consequence is that each card declares one more attack than it actually has, so anything that counts attacks, or pairs them against another data source, sees a card whose moves do not line up.

### Affected sets

| set | cards |
|---|---|
| Unified Minds | 13 |
| Cosmic Eclipse | 7 |
| Unbroken Bonds | 6 |
| Ultra Prism | 4 |
| SM Black Star Promos | 4 |
| Forbidden Light | 3 |
| XY Black Star Promos | 3 |
| Lost Thunder | 2 |
| Team Up | 2 |
| Dragon Majesty | 1 |

Deletions only — 660 lines removed, none added — and `bun run validate` (`tsc --noEmit`) passes locally. A re-scan of `data/` after the change finds no duplicate stubs remaining.

### DIFF
--- a/data/Sun & Moon/Cosmic Eclipse/210.ts
+++ b/data/Sun & Moon/Cosmic Eclipse/210.ts
@@ -95,20 +95,6 @@ const card: Card = {
 			}
 
 		},
-		{
-			cost: [
-				"Colorless",
-				"Colorless",
-				"Colorless",
-			],
-			name: {
-				fr: "Plante Solaire-GX",
-			},
-			effect: {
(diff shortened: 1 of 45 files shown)


Que-tin:
@FalconChipp

PR #1974, opened 2026-08-12
(fix) Remove wrong Pokédex descriptions from Scarlet & Violet ex cards

Que-tin:
### How and why?

While filling the German translation gaps in Scarlet & Violet (sv01) I noticed that 13 cards keep an English `description` that can never be translated: they are all Pokémon **ex** cards, and ex cards don't have a Pokédex entry printed on the physical card. That's also why no other language ever had this text — there is no localised source for it.

Since the text isn't on the card, this PR removes the `description` field from those 13 files instead of translating it:

`019` Spidops ex, `032` Arcanine ex, `045` Gyarados ex, `065` Magnezone ex, `081` Miraidon ex, `086` Gardevoir ex, `088` Banette ex, `123` Great Tusk ex, `125` Koraidon ex, `131` Toxicroak ex, `143` Iron Treads ex, `158` Oinkologne ex, `223` Spidops ex

Only the three-line `description` block is removed in each file; every other byte is unchanged, and `tsc --noEmit` still passes.

### DIFF
--- a/data/Scarlet & Violet/Scarlet & Violet/019.ts
+++ b/data/Scarlet & Violet/Scarlet & Violet/019.ts
@@ -90,9 +90,6 @@ const card: Card = {
 	suffix: "ex",
 	illustrator: "takuyoa",
 
-	description: {
-		en: "The thread it secretes from its rear is as strong as wire. The secret behind the thread's strength is the topic of ongoing research.",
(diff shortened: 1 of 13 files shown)


Que-tin:
@FalconChipp

PR #1992, opened 2026-08-13
spelling and grammer in templates

jswalks:
Correction to grammar and spelling in the issue templates

### DIFF
--- a/.github/ISSUE_TEMPLATE/database-issue.yml
+++ b/.github/ISSUE_TEMPLATE/database-issue.yml
@@ -9,7 +9,7 @@ body:
       value: |
         Thanks for taking the time to help advance the project with your issue report!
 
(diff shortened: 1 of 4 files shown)


.github/ISSUE_TEMPLATE/database-issue.yml
```diff
@@ -21,7 +21,7 @@ body:
       options:
         - Card
         - Set
-        - Serie
```
FalconChipp, line 24: Serie is used API wide. Better to keep as it is rather than correcting at this stage

FalconChipp (changes requested):

.github/ISSUE_TEMPLATE/database-issue.yml
jswalks: Cool, will revert this

FalconChipp (approved):

PR #1993, opened 2026-08-13
[feat]: Attach third-party IDs to pricing objects

FalconChipp:
Include the source third-party IDs (`cardmarket` and `tcgplayer`) from `card.thirdParty` / `variant.thirdParty` into the returned pricing objects. Pricing entries are now `{ id, ...providerData }` or `null` if the provider returned no data. `thirdParty` is not exposed directly.

Card-level (`swsh1-1`)

```json
{
  "cardmarket": {
    "id": 427231,
    "unit": "EUR",
    "avg": 1.23,
    "low": 0.75,
    "trend": 1.1
  },
  "tcgplayer": {
    "id": 206047,
    "unit": "USD",
    "holo": { "marketPrice": 1.1 }
  }
}
```

Variant-level (`swsh4-1`)

```json
[
  {
    "type": "normal",
    "variantId": "endfynwn4n10gzq",
    "pricing": {
      "cardmarket": { "id": 511425, "unit": "EUR", "avg": 0.45 },
      "tcgplayer": { "id": 226360, "unit": "USD", "normal": { "marketPrice": 0.22 } }
    }
  },
  {
    "type": "reverse",
    "variantId": "cm4kqul3x1bwlz1f",
    "pricing": {
      "cardmarket": { "id": 511425, "unit": "EUR", "avg": 0.45 },
      "tcgplayer": { "id": 226360, "unit": "USD", "reverse": { "marketPrice": 0.30 } }
    }
  }
]
```

### DIFF
--- a/server/src/V2/Components/Card.ts
+++ b/server/src/V2/Components/Card.ts
@@ -125,7 +125,12 @@ async function loadCard(lang: SupportedLanguages, id: string): Promise<SDKCard |
 				getCardMarketPrice(variant),
 				getTCGPlayerPrice(variant),
 			]);
-			variant.pricing = { cardmarket, tcgplayer };
+			const cmId = variant.thirdParty.cardmarket
+			const tcgId = variant.thirdParty.tcgplayer
+			variant.pricing = {
+				cardmarket: cardmarket ? { id: cmId, ...cardmarket } : null,
+				tcgplayer: tcgplayer ? { id: tcgId, ...tcgplayer } : null,
+			};
 		}
 	}
 
@@ -148,11 +153,13 @@ async function loadCard(lang: SupportedLanguages, id: string): Promise<SDKCard |
 
 	// console.timeEnd('loading providers')
 	// console.time('remapping card')
+	const cmId = card.thirdParty?.cardmarket
+	const tcgId = card.thirdParty?.tcgplayer
 	const res = {
 		...deepOmit(card, 'thirdParty'),
 		pricing: {
-			cardmarket: cardmarket,
-			tcgplayer: tcgplayer
+			cardmarket: cardmarket ? { id: cmId, ...cardmarket } : null,
+			tcgplayer: tcgplayer ? { id: tcgId, ...tcgplayer } : null,
 		}
 	} as SDKCard
 	// console.timeEnd('remapping card')


server/src/V2/Components/Card.ts
```diff
@@ -125,7 +125,12 @@ async function loadCard(lang: SupportedLanguages, id: string): Promise<SDKCard |
 				getCardMarketPrice(variant),
 				getTCGPlayerPrice(variant),
 			]);
-			variant.pricing = { cardmarket, tcgplayer };
+			const cmId = variant.thirdParty.cardmarket
+			const tcgId = variant.thirdParty.tcgplayer
+			variant.pricing = {
+				cardmarket: cardmarket ? { id: cmId, ...cardmarket } : null,
```
advra, line 131: You should change this line. Currently it does not return ids even though ids for a variant are defined. For example signature variants like set `ex1-5` for `reed-weichler` variant has cardmarketid undefined and  tcgplayerid defined as 477376. We expect cardmarket to be null. But tcgplayer to display the id.

You current implementation for ex1-15 incorrectly shows tcgplayer as null:
```json
    {
      "type": "normal",
      "size": "standard",
      "stamp": [
        "reed-weichler"
      ],
      "variantId": "1t06ikeqvv0boawin32pqrsc4t2ry",
      "pricing": {
        "cardmarket": null,
        "tcgplayer": null
      }
    }
```

I recommend changing lines 131 and 132 to the following:
```javascript
    variant.pricing = {
		cardmarket: cmId ? { id: cmId, ...cardmarket } : { id: null },
		tcgplayer: tcgId ? { id: tcgId, ...tcgplayer } : { id: null },
    };
```
Now the result for ex1-5 will correclty show cardmarket as null and tcgplayerid as 477376: 
```json
    {
      "type": "normal",
      "size": "standard",
      "stamp": [
        "reed-weichler"
      ],
      "variantId": "1t06ikeqvv0boawin32pqrsc4t2ry",
      "pricing": {
        "cardmarket": {
          "id": null
        },
        "tcgplayer": {
          "id": 477376
        }
      }
    }
```

thomas-bassett:
adding a comment based on the discussions we have had privately. 

we shouldn't  have the ids in the pricing objects we should keep them in thirdParty and expose that object, my reasoning would be what if we added thirdParty ids that where not linked to pricing. i.e `ebayIds` `sku` 

i think its also just cleaner to return a null pricing as a hard error thats easy for users to check against

PR #1998, opened 2026-08-13
feat(card): add explicit Tera classification

whosthatbloke:
## Summary

- add an optional `tera` field for an explicitly identified printed Tera Pokémon
- pass the field through the compiler and expose it in REST/OpenAPI and GraphQL
- annotate all 146 qualifying international prints currently represented in the database

This is assertion-only provider data. It is never inferred from a Pokémon's name, artwork, `ex` suffix, or text that merely refers to Tera Pokémon. Consumers can therefore distinguish `tera: true` from an older or unreviewed record where the field is absent.

## Data review

The annotated set/collector-number inventory was reconciled exactly against the current structured PokémonTCG `Tera` subtype inventory: 146 expected, 146 annotated, no extras, and no omissions. Ho-Oh `sv8-019` is a representative close negative: its attack mentions Tera Pokémon, but the card itself is not marked.

## Verification

- `git diff --check`
- root TypeScript validation passed in a clean Bun container
- Docker image built successfully from commit `fc513d26f4c7c8981fc7a7c047a759d0f9af144f`

## Scope

No generated files, lockfiles, Docker configuration, filters, special endpoints, or `data-asia` records are changed.

### DIFF
--- a/data/Mega Evolution/Ascended Heroes/038.ts
+++ b/data/Mega Evolution/Ascended Heroes/038.ts
@@ -31,6 +31,7 @@ const card: Card = {
 	hp: 320,
 	types: ["Fire"],
 	stage: "Stage2",
+	tera: true,
 
 	attacks: [{
 		cost: ["Fire", "Colorless", "Colorless"],
--- a/data/Mega Evolution/Ascended Heroes/057.ts
+++ b/data/Mega Evolution/Ascended Heroes/057.ts
@@ -22,6 +22,7 @@ const card: Card = {
 	hp: 200,
 	types: ["Lightning"],
 	stage: "Basic",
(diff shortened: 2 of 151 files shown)


whosthatbloke:
Closing this draft because the narrow `tera: boolean` design has been superseded.

The replacement uses a canonical multi-valued `traits` array that supports Ancient, Future, Tera, Single Strike, Rapid Strike, and Fusion Strike across Pokémon, Trainer, and Energy cards, including cards with multiple simultaneous traits. The complete 641-card backfill is now merged in the fork.

I will raise a GitHub Discussion for maintainers to agree the broader public contract before proposing a replacement upstream PR.

Issue #1936, opened 2026-08-08
issue: Overview of all cardmarket idProduct used for multiple distinct cards


joachimbaten:
### What seems to be the issue

Other

### Field(s)

pricing:cardmarkt:idProduct

### Description

This table gives a list of all cardmarket idproducts that are used for multiple distinct cards.
If I can somehow help fixing this, point me in the correct direction on how to solve it and I will try to help.

| cardmarket_product_id | assignments | distinct_cards | card_ids |
|---|---:|---:|---|
| 276798 | 10 | 10 | ex11-35, ex11-36, ex11-37, ex11-38, ex11-39, ex11-40, ex11-61, ex11-62, ex11-63, ex11-64 |
| 278861 | 9 | 9 | pl4-AR1, pl4-AR2, pl4-AR3, pl4-AR4, pl4-AR5, pl4-AR6, pl4-AR7, pl4-AR8, pl4-AR9 |
| 274151 | 8 | 8 | gym1-15, gym1-66, gym1-68, gym1-69, gym1-70, gym1-72, gym1-74, gym1-98 |
| 274152 | 7 | 7 | gym1-100, gym1-16, gym1-46, gym1-49, gym1-76, gym1-77, gym1-78 |
| 554217 | 7 | 7 | xyp-XY203, xyp-XY204, xyp-XY205, xyp-XY206, xyp-XY207, xyp-XY208, xyp-XY210 |
| 286307 | 6 | 6 | xy8-157, xy8-158, xy8-163, xy8-164, xy8-61, xy8-62 |
| 276419 | 5 | 5 | ex8-16, ex8-17, ex8-18, ex8-98, ex8-99 |
| 274153 | 4 | 4 | gym1-101, gym1-17, gym1-80, gym1-83 |
| 274154 | 4 | 4 | gym1-102, gym1-18, gym1-57, gym1-85 |
| 274287 | 4 | 4 | gym2-106, gym2-19, gym2-79, gym2-80 |
| 274288 | 4 | 4 | gym2-110, gym2-20, gym2-94, gym2-97 |
| 276973 | 4 | 4 | ex13-3, ex13-4, ex13-5, ex13-6 |
| 278430 | 4 | 4 | pl1-10, pl1-27, pl1-28, pl1-9 |
| 297478 | 4 | 4 | sm2-14, sm2-46, sm2-55, sm2-56 |
[...]

PR #2007, opened 2026-08-13
feat(cardmarket): add audit and safe-fix tooling

joachimbaten:
Provides a script to address the issue #1936 

## Changes

Add a Bun-based Cardmarket audit and fix tool for validating links in the data dataset against Cardmarket’s singles catalog.

## Details

The tool detects missing, invalid, and duplicate Cardmarket links, generates compact audit reports, and proposes conservative fixes. Safe fixes can be applied to selected sets using exact TypeScript AST locations, with collision and cross-set scope checks. The feature also includes progress reporting, documentation, and network-free regression tests. No card-set data changes are included.


FalconChipp:
I have had to close all your PRs relating to cardmarket IDs for a couple reasons. 
1. 82 PRs opened within a matter of minutes, crashing the CI and automated checks
2. Format being used is not the thirdParty block, it's the variants block. 

Please reach out for further questions

joachimbaten:
@FalconChipp

First of all, my apologies for overloading the system. That was never my intention, and I hope it didn't cause too much disruption.

Based on your feedback, I split the original work into separate PRs per set. To do that, I automated the branch and PR creation process, which unfortunately resulted in 84 PRs being opened within a very short timeframe.

In the process, this PR, which contains the tooling, documentation and explanation of the approach, got buried among the set-specific PRs and was also closed.

I still believe many of those PRs contained valuable Cardmarket data fixes. They were generated using the algorithm described in this PR, and I manually validated a significant number of the results. I'd like to understand the preferred way to contribute this kind of large-scale data correction while avoiding unnecessary load on the project and following the expected contribution patterns.

My proposed next steps would be:
1. Update the script so that missing Cardmarket IDs are added inside the variant blocks (alongside tcgplayerId) instead of creating a card-level thirdParty block.
2. For cards that already have a Cardmarket ID in the card-level thirdParty block, only update the existing value if needed, without relocating the data to the variants. This keeps the changeset as small as possible.
3. Create a PR for a single set containing only corrections to existing Cardmarket IDs.
4. Create a separate PR for a single set containing only newly added Cardmarket IDs.

Does this approach align better with the project's expectations, or would you recommend a different contribution strategy?

And again, my apologies for the disruption. I'll be more conservative with future bulk contributions..

PR #2101, opened 2026-08-14
fix(data): Add missing French evolveFrom translations for me05

LoopingStudio:
## Changes

Adds the missing French `evolveFrom` values to the 47 MEP Pitch Black (`me05`) cards that only carried the English one.

## Details

- Each French value is the `name.fr` already used by that Pokémon's own cards elsewhere in the database, so no new translation is introduced — existing ones are only propagated.
- All 47 were cross-checked against the official French species names. The 30 distinct species match exactly; the two Trainer entries (`Antique Skull Fossil` → `Fossile Crâne Ancien`, `Antique Armor Fossil` → `Fossile Armure Ancien`) match cards `072` and `073` of this very set.
- Only the `fr` key is added, no other field is touched: 47 files, +94/-47, two lines each.

This is the first of a series — around 756 other `evolveFrom` entries are still missing their French value across other sets, which I will submit one set per PR.


FalconChipp:
Please address the merge conflicts

PR #2102, opened 2026-08-14
fix(data): Add missing French evolveFrom translations for me04

LoopingStudio:
## Changes

Adds the missing French `evolveFrom` values to the 51 MEP Chaos Rising (`me04`) cards that only carried the English one.

## Details

- Each French value is the `name.fr` already used by that Pokémon's own cards elsewhere in the database, so no new translation is introduced — existing ones are only propagated.
- Every value was cross-checked against the official French species names and matches exactly.
- Only the `fr` key is added, no other field is touched: 51 files, +102/-51, two lines each.

Same shape as #2101 (`me05`), part of a series covering the ~800 `evolveFrom` entries still missing their French value, one set per PR.


FalconChipp:
Please address the merge conflicts

PR #2127, opened 2026-08-14
Fix/1991 third party sm black star

### LINKED ISSUE #1991: issue: third party information missing for all SM black star Promos
### What seems to be the issue

Set

### Field(s)

Third party fields

### Description

No tcgplayer or cardmarket data for any of the SM promos

Will fix this when I get a minute

jswalks:
closes #1991 

## Changes

Added third party IDs for S&M promos

## Details

Added in the IDs for both tcgplayer and cardmarket


data/Sun & Moon/SM Black Star Promos/SM05.ts
```diff
@@ -118,6 +118,11 @@ const card: Card = {
 	],
 
 	retreat: 4,
+
+	thirdParty: {
+		cardmarket: 295297,
+		tcgplayer: 126027
+	},
```
FalconChipp: Needs to be in the variants block rather than standalone thirdParty

FalconChipp (changes requested):

data/Sun & Moon/SM Black Star Promos/SM05.ts
jswalks: I don't quite understand this point, there is no variant block in that example and all other ones I am seeing have it as a stand alone block.

I can see it in like base set 2:

	variants: [
		{
			type: "holo",
			thirdParty: {
				cardmarket: 273924,
				tcgplayer: 42444
			}
		}
	]
	
But like lost thunder have it as I have done, Are we migrating to this variant method?

jswalks: All good adding in

jswalks: sorted

jswalks:
Adding the third-party pricing data (variants/thirdParty blocks) to 249 cards pushed tsc's memory use just past Node's default heap limit on the macOS CI runner, so validate now needs a higher --max-old-space-size to keep passing.

package.json
```diff
@@ -1,7 +1,7 @@
 {
 	"private": true,
 	"scripts": {
-		"validate": "tsc --noEmit --project tsconfig.json",
+		"validate": "node --max-old-space-size=4096 node_modules/typescript/bin/tsc --noEmit --project tsconfig.json",
 		"check-missing": "ts-node meta/scripts/check-missing.ts",
 		"check-missing:help": "echo 'Usage: npm run check-missing \"<pattern>\" \"<property>\" [output-file]'",
 		"link-cardmarket": "ts-node meta/scripts/linkCardToCardmarket.ts",
```
FalconChipp, line 7: Please revert these changes and push. 
Just be careful when pushing with local changes to package.json

FalconChipp (changes requested):

package.json
jswalks: I've had to bump the heap memory size for tsc as we are hitting the default max. it won't change the cost but will improve the performance. Actions runs on clock minutes not memory size anyways. So technically could run faster/cheaper as it's not getting squeezed through GC thrashing. The checks were failing due to memory issues without this change

jswalks: Do we just split the PRs then if we dont adjust the heap memory?

FalconChipp: Yeah for now. Its not our call to make as maintainers so if you'd like to, open a separate PR with the heap bump and Avior can make the call

jswalks: Okay nw. I'll close this and split into 2 PRs

FalconChipp (approved):

PR #2128, opened 2026-08-14
Fix: OBF Correct Cardnumbers and add Cardmarket Ids and TCGPlayer IDs direct t…

MineTommy:
## Changes

Fix any Cardmarket Ids and adding Variants (Sometimes there are switched with Cosmo / Normal


FalconChipp:
Nice work!

PR #1810, opened 2026-07-23
(data) Dragon ex3 Added Missing Variants

advra:
closes # N/A

## Changes

- added missing card variant data

## Details

added all missing card variant data for ex dragon (ex3)

PR #2171, opened 2026-08-16
(fix) Remove duplicated attacks from Skyridge Trainer cards

Que-tin:
### What

Six Trainer cards in Skyridge carry an `attacks` entry that should not exist:

- `132` Mirage Stadium (Stadium)
- `137` Mystery Zone (Stadium)
- `138` Oracle (Supporter)
- `139` Star Piece (Tool)
- `140` Underground Expedition (Supporter)
- `141` Underground Lake (Stadium)

On each of them the attack has an empty name (`name: {}`, marked `// name intentionally left blank`) and an English effect that is a verbatim copy of (part of) the card's own Trainer `effect`.

### Where it came from

The blocks arrived with the bulk variants import in #1810. A scan over the whole repository (attack effect text contained verbatim in the Trainer effect) finds exactly these six files, so Skyridge is the full extent of it.


LoopingStudio:
Hey :), how do you make the "Test the Data" actions to run ? mine are "Waiting for the status to be reported"

Issue #1612, opened 2026-06-07
issue:cel25-17 falsely has an abillity


internetyay:
### What seems to be the issue

Card

### Field(s)

card database

### Description

The card at this location: data/Sword & Shield/Celebrations/17.ts should not have an ability. In the file, it has one in french

viet456:
Resolved in #2183 , this may be closed now.

PR #2183, opened 2026-08-16
(fix) Remove fr-only stray ability from Groudon (cel25-17)

Que-tin:
Groudon (Celebrations 17) prints no ability — it is a Basic Fighting Pokémon with two attacks, Magma Volcano and Massive Rend. Its file nevertheless carries an `abilities` entry with French content only:

```ts
type:      "Poke-POWER"
name.fr:   "Rayon Obscur"
effect.fr: "… lorsque vous placez Noctali Star de votre main sur votre Banc …"
```

OCR of the German print (017/025) reads `BASIS / Groudon / KP 130 / Magmavulkan / Riesiger Riss` and no ability box.

Same defect class as #2136. `npm run validate` passes.

### DIFF
--- a/data/Sword & Shield/Celebrations/17.ts
+++ b/data/Sword & Shield/Celebrations/17.ts
@@ -21,18 +21,6 @@ const card: Card = {
 	hp: 130,
 	stage: "Basic",
 
-	abilities: [
-		{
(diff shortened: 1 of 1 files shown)


viet456:
Resolves #1612

PR #2184, opened 2026-08-16
Remove the `thirdParty` deepOmit

FalconChipp:
## Changes

Remove the `deepOmit` stripping out thirdParty IDs from card responses. 

## Details 

Only change to this for V2, 
Rework #1993 for V3

### DIFF
--- a/server/src/V2/Components/Card.ts
+++ b/server/src/V2/Components/Card.ts
@@ -149,7 +149,7 @@ async function loadCard(lang: SupportedLanguages, id: string): Promise<SDKCard |
 	// console.timeEnd('loading providers')
 	// console.time('remapping card')
 	const res = {
-		...deepOmit(card, 'thirdParty'),
+		...card,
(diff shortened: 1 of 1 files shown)


advra:
Thank you for this!

thomas-bassett (dismissed):

server/src/V2/Components/Card.ts
```diff
@@ -22,7 +22,7 @@ import zhtw from '../../../generated/zh-tw/cards.json'
 import { getCardMarketPrice } from '../../libs/providers/cardmarket'
 import { getTCGPlayerPrice } from '../../libs/providers/tcgplayer'
 import { executeQuery, type Query } from '../../libs/QueryEngine/filter'
-import { deepOmit } from "../../util";
+// import { deepOmit } from "../../util";
```
thomas-bassett: can you remove this comment if we don't need it

PR #2194, opened 2026-08-17
[feat]: New status page

FalconChipp:
## Changes

Rewrite `/status` and add supporting endpoints:

- **Dynamic GitHub SVG badge** — layout and columns generated from whichever languages are compiled in; automatically adds a Total Asia column when Asian data is present; includes CardMarket and TCGPlayer pricing coverage rows
- **Asia data endpoint** — new `GET /status/data/asia/:lang` returns per-series/set breakdown for Asian languages (ja, ko, zh-tw, zh-cn, id, th), which use completely different set IDs from the international catalogue
- **Dataset version endpoint** — new `GET /status/data/version` returns version, commit, generatedAt, and schema as a lightweight payload
- **Revamped status UI** — matches tcgdex.dev styling (Lexend font, `#c80040` accent, dark-mode aware CSS variables); collapsible sections; Set Completion table with Summary / International / Asia filter tabs; per-set card drilldown modal; language codes shown alongside names
- **Header pills** — server uptime, dataset version/commit (linked to this repo), generated date, schema version, and link to status.tcgdex.dev

## Details

Asian languages (ja, ko, zh-tw, zh-cn, id, th) cannot be represented in the international Sets table because they use independent set hierarchies. The `/data/asia/:lang` endpoint exposes their per-series/set counts, and the status page renders them in a dedicated view with a language picker and lazy-loaded series accordion.

The GitHub SVG badge is fully dynamic: columns are built from `langStats` at request time (sorted alphabetically), so adding or removing a compiled language requires no code changes.

Git version and commit hash are baked into `generated/git-info.json` during the Docker build stage (where git is already installed), so the version pill in the header is always accurate with no manual env var updates needed. The commit timestamp (`git log -1 --format=%cI`) is also captured as `generatedAt`.

The status module now stores only three booleans per card (`image`, `hasCm`, `hasTp`) instead of full variant arrays, keeping per-worker memory footprint small.


Dockerfile
```diff
@@ -23,7 +23,11 @@ ADD --chown=bun:bun . .
 
 # build
 RUN cd server && \
-	bun run compile
+	bun run compile && \
+	printf '{"version":"%s","commit":"%s"}' \
```
Aviortheking: it would be better to add a hook into the start of the server, also the Dockerfile.github-actions doesn't include this change :D

thomas-bassett (changes requested):
Can we expose the github commit/release as an endpoint. I would also say add the time the commit/release happened too if not already there

advra:
What do you think about adding a dataset section and moving the version/commit down into it? Maybe include a schema version and timestamp for when it was generated could be helpful 

```json
"dataset": {
  "generatedAt": "2026-09-06T12:31:42Z",
  "version": "2026.09.06",
  "commit": "31556f0",
  "schema": "v2"
}
```

FalconChipp:
@thomas-bassett & @advra 

image

and at "/status/data"
```
"dataset": {
    "version": "v2.47.0",
    "commit": "df9a30c",
    "generatedAt": "2026-08-31T20:57:35+01:00",
    "schema": "v2"
  },
  ```

thomas-bassett (approved):

PR #2197, opened 2026-08-17
feat(api): expose updated timestamp on card stub/brief shapes

FrostyFitz:
`CardResume` (the stub shape used for cards nested inside `/sets/:id`
and `/series/:id` responses, and for the flat `/cards` listing) doesn't
carry an `updated` field, even though it's already computed via
`getCardLastEdit()` for the full single-card shape.
This makes incremental sync effectively impossible for API consumers —
there's no way to tell which cards changed without re-fetching every
card individually, or the entire catalog on every sync.
This PR just plumbs the already-computed value through:
- `cardToCardSimple()` (used when compiling nested Set/Series card
  lists) now includes `updated`, same as its sibling `cardToCardSingle()`
  already does for full card fetches.
- `toBrief()` (used for the `/cards` listing endpoint) now includes
  `updated`, since the `SDKCard` it derives from already carries the
  field.
- `CardResume` gains an optional `updated?: string` field.
No new data collection, no breaking change (field is additive and
optional), no change to existing full single-card response shape.
Happy to adjust scope/direction if you'd rather this land differently.


thomas-bassett (changes requested):
Thanks for the PR. I want to push back on it just a little and maybe steer in a better direction. instead of increasing the response of the briefs and requiring more checks for users can expand the filtering behaviour instead. at the moment we have a [Greater Than](https://tcgdex.dev/rest/filtering-sorting-pagination#greater-than-gt)  but that only supports numbers. we should expand the support to include datetime stamps aswell. this would allow a user to call `https://api.tcgdex.net/v2/en/cards?updated=gte:2026-08-19T08:25:49+01:00` and get a list of all the cards that have been updated, without the need to go through every set.

thomas-bassett:
will be fixed in `v3`

PR #2200, opened 2026-08-18
fix(data): Chaos Rising variant and marketplace id fixes

Sinrath:
013 was missing the reverse holo, 029 had holo/reverse on the wrong product and 068 had a duplicated holo/reverse pair. The Build & Battle non-holos (013, 029, 051, 068, 085) now point to their own Cardmarket and TCGplayer products instead of the pack cards. All ids checked against the Cardmarket product list / price guide and the TCGplayer API.


MineTommy:
Thanks for fixing.

I have all 5 Cards as Issues on my Table ;-)

PR #2284, opened 2026-08-26
feat(data): add the zh-cn card list for CSV9C 星彩晶璃

CorruptFun:
`data-asia/SV/CSV9C.ts` has existed as a shell since the set was added, so
`/v2/zh-cn/sets/CSV9C` answers `cards: []` and a Simplified-Chinese card from
星彩晶璃 cannot be identified by anything reading the API. This adds the 264
card files.

All 57 `zh-cn` sets are in the same state — set metadata, zero card rows — so
this is one set offered as a **proposal about shape** rather than a bulk drop.
I have the same data prepared for 78 Simplified-Chinese sets / ~9,250 cards and
would rather agree the format on one set than open a PR with thousands of files
you did not ask for. Tell me what you would like changed and I will regenerate
the lot.

## Shape

Follows `data-asia/S/SC2D`, the populated Traditional-Chinese set already in the
tree: a single-language `name`, a `category`, nothing inferred.

```ts
import { Card } from "../../../interfaces"
import Set from "../CSV9C"

const card: Card = {
	set: Set,

	name: {
		'zh-cn': "太乐巴戈斯"
	},

	category: "Pokemon",
	rarity: "Rare",
}

export default card
```

`rarity` is mapped conservatively from the Chinese rarity letters
(C/U/R/RR/AR/SR/SAR/UR/ACE → Common/Uncommon/Rare/Double rare/Illustration
rare/Ultra Rare/Special illustration rare/Hyper rare/ACE SPEC Rare). Letters
with no unambiguous counterpart are **omitted** rather than approximated —
`SC2D`'s cards carry no rarity at all, so an absent rarity is a shape this
repository already contains.

`trainerType` and `energyType` come from the card-type column
(物品卡 → Item, 支援者卡 → Supporter, 竞技场卡 → Stadium, 宝可梦道具 → Tool;
基本…能量 → Normal, otherwise Special).

## What is deliberately missing

`hp`, `types`, `attacks`, `description`, `illustrator`, `dexId`, `retreat`,
`weaknesses`. The source is a **checklist**, so none of them are known, and
every one would have to be guessed or carried over from the Japanese card this
printing reprints. I would be glad to do that as follow-up — every row carries
the JP set it was reprinted from, so the join is available — but not as an
inference dressed as data.

`name` carries `zh-cn` only. The official English species name is known for
most of these cards and is **not** written into the name table: these printings
have no English release, and an English name on a Chinese-exclusive card is how
a Chinese card starts being answered by an English one.

## Two cards are missing on purpose

The source lists **136/208 twice** — as 铝钢龙 (Duraludon, C) and as 铝钢桥龙
(Archaludon, R holo) — while **137/208 is absent**. They are adjacent
evolutions and 138 is 铝钢桥龙ex, so the intended numbering is obvious to a
reader; that is exactly why it is not encoded here. Say the word and I will add
136 = 铝钢龙 and 137 = 铝钢桥龙, or I can raise it upstream first.

## Provenance, and a licensing question I would rather ask than assume

The data is extracted from **神奇宝贝百科 (wiki.52poke.com)** via the MediaWiki
API — wikitext, not rendered HTML — and cross-checked against your own
`/v2/zh-cn/sets` list. Extraction is polite (serialised, 700 ms floor,
identifying User-Agent, everything cached).

That wiki publishes under **CC BY-NC-SA 3.0**. What is taken here is
factual — a set code, a printed number, the name printed on the card, a rarity
letter — with no prose, images or commentary, and facts are not copyrightable
in the US (*Feist*). But you maintain an MIT-licensed database and the
**EU/UK sui generis database right** protects substantial extraction regardless
of originality, so this is your call rather than mine. If you would prefer a
different provenance I am happy to verify the same rows against The Pokémon
Company's own `pokemon.cn` product pages (each set links one) or against
physical cards, and resubmit.

## Verification

- 264 files typecheck against `interfaces.d.ts` with **0 errors**.
- Set completeness is proven before publishing: 1…208 each present exactly once
  (which is what caught the 136/137 error), and the total agreeing with the
  set's own declared `208+58`.
- Spot-checked against photographs of real cards: 174/208 太乐巴戈斯 (Rare,
  holo) and 214/208 猫头夜鹰 (Illustration rare) are two cards a collector sent
  us that no catalogue could identify — which is what prompted the work.

The extractor is open and reusable if you would like it:
<https://github.com/CorruptFun/Cardstock/tree/main/scripts/zh-pokemon>

### DIFF
--- a/data-asia/SV/CSV9C/001.ts
+++ b/data-asia/SV/CSV9C/001.ts
@@ -1,0 +1,15 @@
+import { Card } from "../../../interfaces"
+import Set from "../CSV9C"
+
+const card: Card = {
+	set: Set,
+
+	name: {
+		'zh-cn': "蛋蛋"
+	},
+
+	category: "Pokemon",
+	rarity: "Common",
+}
+
+export default card
--- a/data-asia/SV/CSV9C/002.ts
+++ b/data-asia/SV/CSV9C/002.ts
@@ -1,0 +1,15 @@
+import { Card } from "../../../interfaces"
+import Set from "../CSV9C"
+
+const card: Card = {
+	set: Set,
+
+	name: {
+		'zh-cn': "触手百合"
+	},
+
+	category: "Pokemon",
+	rarity: "Common",
+}
+
+export default card
--- a/data-asia/SV/CSV9C/003.ts
+++ b/data-asia/SV/CSV9C/003.ts
@@ -1,0 +1,15 @@
+import { Card } from "../../../interfaces"
+import Set from "../CSV9C"
+
+const card: Card = {
+	set: Set,
+
+	name: {
+		'zh-cn': "摇篮百合"
+	},
+
+	category: "Pokemon",
+	rarity: "Uncommon",
+}
+
+export default card
--- a/data-asia/SV/CSV9C/004.ts
+++ b/data-asia/SV/CSV9C/004.ts
@@ -1,0 +1,15 @@
+import { Card } from "../../../interfaces"
+import Set from "../CSV9C"
+
+const card: Card = {
+	set: Set,
+
+	name: {
+		'zh-cn': "尖牙笼"
+	},
+
+	category: "Pokemon",
+	rarity: "Common",
+}
+
+export default card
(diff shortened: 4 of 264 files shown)


thomas-bassett:
Thanks for your first PR, couple things we dont allow AI prs in this repository. i can see youve linked a repo that generated this data, i assume that you created that with AI and then used it. please come talk to us in the discord about this.

PR #2285, opened 2026-08-27
fix(data): White Flare and Black Bolt variant and marketplace id fixes

Sinrath:
White Flare
- 018, 040, 051, 054, 062, 078 (Rare): added the plain `reverse` variant, same ids as the holo, like the other Rares of the set.
- 043 Gothitelle (Rare): `normal` changed to `holo`, like the other Rares of the set.
- 011 Tepig: removed a plain `holo`. A Common has no holo print; the promo is already there as `holo/tinsel`.

Black Bolt
- 012 Victini, 070 Haxorus (Rare): `normal` changed to `holo`, like the other Rares of the set.

Set-logo promos
- TCGplayer ids added: Victini BLK 012 (668956), Reshiram ex WHT 020 (668958), Zoroark WHT 062 (668959).
- Zekrom ex BLK 034: stamped variant added (Cardmarket 867786, TCGplayer 668957).


data/Scarlet & Violet/White Flare/011.ts
```diff
@@ -70,13 +70,6 @@ const card: Card = {
 				tcgplayer: 642126
 			}
 		},
-		{
-			type: "holo",
-			thirdParty: {
-				cardmarket: 835923,
-				tcgplayer: 642126
```
thomas-bassett, line 77: lets move tgcplayer down to the tinsel as i belive that is the correct id

thomas-bassett (changes requested):

data/Scarlet & Violet/White Flare/011.ts
Sinrath: Moved 642126 to the tinsel holo variant in db92836.
thanks

thomas-bassett (approved):

PR #2291, opened 2026-08-28
fix(data): Mark Dark Explorers Secret Rares as holo

pggpgg:
## Changes

Dark Explorers Secret Rares 109-111 currently export as normal. This is because cards with a missing cariant always become normal: true, holo: false

In reality, all three secret rares in this set are holofoil-only printings.

## Details

- `109.ts` Gardevoir
- `110.ts` Archeops
- `111.ts` Pokémon Catcher

Each card gets a single `type: "holo"` variant.

## Sources (visual evidence)
- [Gardevoir](https://preview.redd.it/is-this-gardevoir-dark-explorers-card-real-v0-fozvbm8tqkve1.jpg?width=811&format=pjpg&auto=webp&s=a14c38cf2ad8a111c30e38e826c58e57f6226b21)
- [Archeops](https://i.redd.it/gewbvpee95471.jpg)
- [Pokémon Catcher](https://i.ebayimg.com/images/g/zf8AAeSwDfRpDwMT/s-l1600.webp)

### DIFF
--- a/data/Black & White/Dark Explorers/109.ts
+++ b/data/Black & White/Dark Explorers/109.ts
@@ -93,6 +93,12 @@ const card: Card = {
 		de: "Dieses sehr seltene Pokémon hat eine andere Farbe als normal. Es ist sehr schwer zu finden."
 	},
 
+	variants: [
+		{
+			type: "holo"
+		}
+	],
+
 	thirdParty: {
 		cardmarket: 280437,
 		tcgplayer: 85638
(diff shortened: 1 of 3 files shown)


pggpgg:
If this is deemed acceptable by maintainers I'll make a few more related to this set. There are a certain number of cards in this set (namely EX and full-art cards) that end up being computed as normal instead of holo variants.

thomas-bassett:
you can make PRs bigger than this one. we say limit PRs to mainly one set so if its adding variants to every card in a single set thats fine!

it's only when making changes to many cards over many different sets

just keep prs to 1 job sized responsibly and you'll be all good!

Issue #1958, opened 2026-08-11
Asian S-P promo numbers can identify different cards by market


pggpgg:
## Summary

The shared `data-asia/S/S-P/<localId>.ts` representation currently assumes that one `S-P` local number identifies the same physical card in every Asian language. That assumption does not hold for regional Sword & Shield promo lists.

This produces two separate problems:

1. one file can merge metadata from physically unrelated cards; and
2. legitimate market-specific cards cannot be represented at their printed number without colliding with another market.

I searched the existing issue/PR history and found the merged Japanese S-P import, but not an issue covering this cross-market identity collision.[11]

### `135/S-P`

The current `135.ts` combines:

- Japanese **Galarian Zigzagoon** (`ガラル ジグザグマ`), and
- Traditional Chinese **Scorbunny** (`炎兔兒`).[2]

The shared mechanics, type, weakness, `dexId` 263, and most attacks still describe Galarian Zigzagoon, so the Traditional Chinese name/flavor/attack fragments do not belong to the same card object.[2]

Separately, the Thai S-P list identifies **Archeops** as **135/S-P**, distributed in the Paradigm Trigger booster-pack purchase campaign.[6] That physical card cannot be added to the current `135.ts` without introducing a third unrelated identity.

### `190/S-P`

The current `190.ts` combines:

- Japanese **Polteageist** (`ポットデス`), and
- Traditional Chinese **Eelektrik** (`麻麻鰻`).[3]

The shared mechanics and `dexId` 855 describe Polteageist, not Eelektrik.[3]

The Korean S-P list identifies **Archeops** as **190/S-P**, from the Find the Legendary Pokémon! Lugia Get Challenge.[5] It therefore collides with both identities already placed in the shared file.

### Correct Japanese record is unreachable

`327.ts` is a coherent Japanese Archeops record.[4] However, the shared S-P set definition currently declares only Indonesian and Thai names/release dates, not Japanese, Korean, or Traditional Chinese locale metadata.[1] As checked on 2026-08-11, the live Japanese endpoint for `S-P-327` returns 404.[10]

The corresponding Korean `S-P-190` and Thai `S-P-135` endpoints also return 404.[8][9]

## Current behavior

- Files are keyed only by `S-P` plus local number.
- Different regional promo sequences are merged when their printed numbers coincide.
- Card-wide properties such as `dexId`, type, HP, stage, attacks, weakness, illustrator, and variants cannot be correct for every language when the identities differ.
[...]

PR #2294, opened 2026-08-29
Support for shared ids across langs

thomas-bassett:
Closes:
#1958 

## Changes

This pr is to add support for cards that share ids. example used in this case was SV-P-173
in JA and KO they are completely different cards even though a share a card and set id.  There are other cards that share a precedence with this issue especially from old sets

it also adds language specific card counts as sets can now have more or less cards within the same set, backwards compatibility has been added as a "default for all langs" mode as this will still be useful for sets that are aligned. 

## Details
It works by allowing a directory to take the place of the card number. in that directory you can create as many ts files as you like as long as it only defines one card per langauge.
The compiler has been changed to work with these new directories and has some checks in place to raise errors when potential overlaps have been created eg. a single ts file and a directory of the same number, or a directory with two files defining a `ja` card

as an example i have added one card `SV-P-173` in place of a single ts file we now have `173/ditto` && `173/noctowl` 

I did think about instead of directories using different "export" within the same file. I think this while adding more compiler complexity might be better for the open source project to manage as its clear what is going on. it likely may also fit in better with how the web ui editor works... although i guess this might be a breaking change for that.


thomas-bassett:
will be added in `v3`

PR #2297, opened 2026-08-30
fix(data): Correct Plasma Blast Archen-line variants

pggpgg:
## Changes

Plasma Blast Archen 53, Archeops 54, and Plume Fossil 82 currently export as normal-only. That is wrong for Archeops 54 (Rare Holo) and omits reverse holos on all three.

International print texts have also been updated for a handful of languages

Marketplace IDs moved to variant level

## Details

- `53.ts` Archen (Uncommon): `normal` + `reverse`
- `54.ts` Archeops (Rare Holo pack card): `holo` + `reverse`
- `82.ts` Plume Fossil (Uncommon): `normal` + `reverse`

## Portuguese reverse holos

Portuguese Plasma Blast was printed, but **that Portuguese set had no reverse holos**. English, German, French, and Italian did.

This PR still adds unscoped `reverse` on all three cards. The compiler does not language-filter variants, so Portuguese will also export `reverse: true`. That is a known overstatement. A previous attempt to scope reverse to `en`/`fr`/`it`/`de` was rejected, so this follows the accepted unscoped shape. No obvious compiler-compliant way to do this as of right now.

Maybe worthy of an Issue? I currently don't know of a way to override/amend the language list at the variant-level

### DIFF
--- a/data/Black & White/Plasma Blast/53.ts
+++ b/data/Black & White/Plasma Blast/53.ts
@@ -83,20 +83,17 @@ const card: Card = {
 
 	variants: [
 		{
-			type: "normal",
-			thirdParty: {
-				cardmarket: 281074,
-				tcgplayer: 83608
-			}
+			type: "normal"
 		},
 		{
-			type: "reverse",
-			thirdParty: {
-				cardmarket: 281074,
-				tcgplayer: 83608
-			}
+			type: "reverse"
 		}
-	]
+	],
+
+	thirdParty: {
+		cardmarket: 281074,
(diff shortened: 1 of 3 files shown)


thomas-bassett (changes requested):
we should still move the thirdparty blocks into the variants please!

pggpgg:
Done in the latest commit. Moved the Cardmarket and TCGplayer IDs into each normal, holo, and reverse variant.

thomas-bassett (approved):

PR #2314, opened 2026-09-03
fix(data): Replace the English type names left in Rising Rivals French text

LoopingStudio:
> **Stacked on #2313.** Eight of these strings also carry a lost SP suffix, so the two changes touch the same lines and cannot sit on separate bases. This PR contains #2313's commit as well; once that merges, the diff here reduces to the second commit alone.

## Changes

Where the French rules text of Rising Rivals (`pl2`) should carry an energy symbol, it carries the **English type word** instead.

```
en: "Discard a Fire Energy attached to Arcanine. ..."
fr: "Défaussez une Énergie Fire attachée à Arcanin. ..."
de: "Lege 1 {R}-Energie, die an Arkani angelegt ist, auf deinen Ablagestapel. ..."
```

**44 symbols** replacing **35 English type words** across **28 cards**. Only `fr` values change — 32 lines in, 32 lines out. No English type word remains in any French string in the set.

## Sources

The German is the strong signal here: it carries the `{X}` token in the same position in **34 of the 35** cases, which is what makes this a transcription slip rather than a deliberate wording choice. The English names the type in all 35.

The one exception is `36` Kaimorse, whose German lost its token too (`1 deiner -Energiekarten`), so that one rests on the English `Water Energy cards` alone.

## Details

Two strings are not a single type.

`67` Kecleon lists all nine at once — `Kecleon est de type GrassFireWaterLightningPsychicFightingDarknessMetalColorless` — and the German gives the exact sequence, `{G}{R}{W}{L}{P}{F}{D}{M}{C}`.

`102` Énergie Sup carries two, the second a concatenated pair: `fournit de l'Énergie ColorlessColorless` against the German `liefert {C}{C}-Energie`. A word-boundary search misses that one, since the two words run together.

`99` Énergie Obscurité and `100` Énergie Métal are worth noting for the opposite reason: their card *names* are already correctly translated, and only the inline type references inside the rules text stayed English — `n'est pas Darkness`, `fournit de l'Énergie Metal`.

The five Rotom cards `RT1` to `RT5` each state a type change in prose (`est de type Water jusqu'à la fin de votre tour`), all confirmed by the German.

`npm run validate` passes. No English or German string was modified.

### DIFF
--- a/data/Platinum/Rising Rivals/1.ts
+++ b/data/Platinum/Rising Rivals/1.ts
@@ -39,7 +39,7 @@ const card: Card = {
 			},
 			effect: {
 				en: "As long as Arcanine has any Fire Energy attached to it, Arcanine has no Weakness.",
-				fr: "Tant qu'Arcanin possède de l'Énergie Fire, il ne possède pas de Faiblesse.",
+				fr: "Tant qu'Arcanin possède de l'Énergie {R}, il ne possède pas de Faiblesse.",
 				de: "Solange an Arkani mindestens 1 {R}-Energie angelegt ist, hat Arkani keine Schwäche."
 			}
 		},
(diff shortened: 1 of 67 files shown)


LoopingStudio:
Folding this into #2313 instead, so that Rising Rivals is one PR with the two fixes as two separate commits. Eight strings carry both defects on the same line, so the two changes cannot be reviewed from separate bases, and stacking them from a fork makes GitHub render the base PR's commit twice. Nothing is dropped — the commit moves to #2313 unchanged.

PR #2331, opened 2026-09-06
fix(data): Add player-rewards-program variant of swsh7/8

cloetensbrecht:
closes # 

## Changes

Add the player-rewards-program variants for swsh7/8.

## Details

swsh7/8 (Leafeon VMAX 008/203) is included in both Play! Pokémon Prize Pack Series 1 and Series 2.


thomas-bassett (changes requested):
This was a use case that hadn't been thought of when I created the variantId generator. currently this would generate two identical ids so i need to figure that out first before merging.

Marking as `request changes` so others don't merge this

PR #2332, opened 2026-09-06
fix(data): Add player-rewards-program variant of swsh7/7

cloetensbrecht:
closes # 

## Changes

Add the player-rewards-program variants for swsh7/7.

## Details

swsh7/7 (Leafeon V 007/203) is included in both Play! Pokémon Prize Pack Series 1 and Series 2.


thomas-bassett (changes requested):
same issue as #2331

PR #2333, opened 2026-09-06
fix(data): Add holo player-rewards-program stamp variants of swsh7/30

cloetensbrecht:
closes # 

## Changes

Add the player-rewards-program variants for swsh7/30.

## Details

swsh7/30 (Vaporeon VMAX 030/203) is included in both Play! Pokémon Prize Pack Series 1 and Series 2.

### DIFF
--- a/data/Sword & Shield/Evolving Skies/30.ts
+++ b/data/Sword & Shield/Evolving Skies/30.ts
@@ -91,6 +91,22 @@ const card: Card = {
 				tcgplayer: 246736
 			}
 		},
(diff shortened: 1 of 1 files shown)


thomas-bassett (changes requested):
same issue as #2331

thomas-bassett:
also just an fyi your last 3 prs could be merged into a single one. we normally say one PR per set unless you are making a small change on 1-2 cards in each set then you can do by series instead.

PR #2340, opened 2026-09-07
Rename Base set cards to have padded 3 digit names

FalconChipp:
## Changes

Renames all base set cards to follow the 3 digit (leading zeros if needed) format


thomas-bassett:
is it worth writing a wrapper around the codeshift? or move the code shift over to `ts-morph`? not sure how many of these exsist but automating it might be worth the dev time

https://ts-morph.com/

would just mean we could run it after updating the data to the newest commits. save on potential merge conflicts

FalconChipp:
> is it worth writing a wrapper around the codeshift? or move the code shift over to `ts-morph`? not sure how many of these exsist but automating it might be worth the dev time
> 
> https://ts-morph.com/

Yeah thats what ill be doing next. First one was manual to see how long it would take manually and it wasn't worth continuing. I just have to do some reading first

PR #2356, opened 2026-09-14
(data) Add Japanese M set M6a 30th Celebration

KlausDerKleber:
## What

Adds the Japanese set **M6a 拡張パック 30th CELEBRATION**, released 2026-09-16.

- `data-asia/M/M6a.ts` — set definition, 103 official cards
- `data-asia/M/M6a/001.ts` … `165.ts` — the printed set (001–103), the alt-art secrets
  (104–135) and the 30 Classic Collection reprints (136–165)
- `data-asia/M/M6a/R.ts`, `G.ts`, `B.ts` — the three RGB Rare Mew prints
- `data-asia/M/M6a/GRA.ts` … `MET.ts` — the eight basic energies
- `interfaces.d.ts` — adds the two Japanese rarities this set introduces, **`Futuristic Rare`**
  and **`RGB Rare`**, the same way SM11b added `Character Rare` / `Character Super Rare`

All **176** cards of the set.

## Data sources

- **pokemon-card.com** (official): Japanese names, card text, flavor text, dex numbers,
  illustrators — and, for the Classic Collection, the full card structure
- **limitlesstcg.com**: stage, evolves-from, weakness/resistance/retreat, regulation mark,
  the ten RR and the basic energies
- **Bulbapedia** (`30th Celebration (TCG)`, Japanese section): the secret rarities (AR, SAR,
  FUR), the Classic Collection's regulation marks, and the RGB Mews
- **serebii**: secret-rare illustrators
- **Cardmarket**: `thirdParty.cardmarket` on the variants, from expansion 6603

## Four things about this set are unusual

**93 of the 103 printed cards carry `rarity: "None"`.** Not missing data — the set
reproduces classic card faces and only the ten ex cards print a rarity symbol (RR). Verified
on the scans: 001 and 101 read `001/103` and `101/103` with nothing after the number, while
009 reads `009/103 RR`. limitless and Bulbapedia mark exactly the same ten.

**136–165 are the set's Classic Collection** — 30 gold-bordered reprints from Base Set
onward, carrying the existing `Classic Collection` rarity. No structural source had them:
limitless has no M6a pages for them and its Japanese database does not reach their original
prints (エリカのプリン returns nothing at all). They are therefore read from the official
card page. That parser is held against limitless on every card where both sources exist —
across this set's other 131 cards it reproduces limitless exactly, with one disagreement
noted below. Their regulation marks follow the original print (none before 2019, then A to
G); the official page shows only the set logo, so those come from Bulbapedia, spot-checked
[...]

PR #2361, opened 2026-09-16
(data) Add English sets 30c 30th Celebration and its Classic Collection

KlausDerKleber:
> English only for now — the other five languages are not available from any source this was built from. Everything else for both sets is complete.

## What

Adds the western release of **30th Celebration**, out worldwide on 2026-09-16 alongside the Japanese M6a.

- `data/Mega Evolution/30th Celebration.ts` — set definition, id `30c`, 128 official cards
- `data/Mega Evolution/30th Celebration/001.ts` … `158.ts` — the printed set and its secrets
- `data/Mega Evolution/30th Celebration/R.ts`, `G.ts`, `B.ts` — the three RGB Rare Mew prints,
  which reprint 065
- `interfaces.d.ts` — adds the three rarities this set introduces, **`Futuristic Rare`** (157/158),
  **`Pikachu Rare`** (the set's thirty Pikachu cards) and **`RGB Rare`**

161 cards in all.

## Data sources

- **Bulbapedia card pages** (`<Name> (30th Celebration <number>)`): stage, type, HP, attacks with
  cost, damage and effect, abilities, weakness, resistance, retreat, illustrator, national dex
  and flavour text
- **TCGplayer via tcgcsv** (group 24722): rarity and `thirdParty.tcgplayer`
- **Cardmarket** expansion 6601: `thirdParty.cardmarket`

Bulbapedia's set list supplies the card list itself: it has all 158 numbers, four of which
(143, 151, 152, 154) TCGplayer has not listed yet.

## Cardmarket ids are aligned, not counted

Expansion 6601 covers this set and its Classic Collection in one, and its product order is **not**
the collection order: Ninetales (010) sits between 128 and 129, one id is skipped outright
(907626), and four cards were added days later. Counting positions would have mis-assigned ids
from card 10 onward. They are therefore matched by product name against the card list, and each of
the 158 numbers resolves to exactly one product no other card claims.

## An independent check

Bulbapedia states each card's Japanese counterpart (`jpcardno`). Across the **103 cards this set
shares with M6a**, HP, retreat cost, weakness and type agree on **every one** — two sets built
from different sources, matching field for field. The Japanese side is #2356.

## The Classic Collection

Its own set (`30ccc`), as TCGplayer lists it and as tcgdex models Celebrations 2021. These 30 cards
reproduce earlier cards down to their original set's collection number — `4/102` Charizard from
Base Set, `203/193` Magikarp from Paldea Evolved — and this repository already holds every one of
those originals. Each card is therefore **derived from the original's own file**: everything that
describes the card is carried over unchanged, including the languages those files already have,
and only what belongs to this printing is replaced — the set it points at, the rarity, the
regulation mark and the marketplace ids. The originals' own `thirdParty` ids and variant blocks are
dropped, since they describe the earlier print.

Per the convention suggested on Discord, localId is the printed number with a letter suffix where
two cards share one, ordered by the original sets' release order: **11a** Metagross δ /
**11b** Genesect-EX, and **106a** Shining Celebi / **106b** Palkia LV.X / **106c** M Gardevoir-EX.

## Not in this PR

**The other five languages for the main set.** Recent sets in this tree carry de/en/es/fr/it/pt on every card; only
English is available from the sources above.

**Reverse-holo variants**, which no source states. Each card carries one variant, holo or normal
by rarity.

`Futuristic Rare` is also added by #2356; whichever lands second will need that one line resolved.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


louisiaegerv (commented):
thank you!

thomas-bassett:
This has been added now but you do have the detailed variants where the current dose not. Im going to close this but could you add the variants to the current main

PR #2327, opened 2026-09-05
fix(data): complete dexId for Unleashed LEGEND cards

lamplis:
## Summary
- Unleashed LEGEND cards 90–95 each name two legendary beasts, but `dexId` only stored one of them.
- Both halves of each pair now list both national dex IDs, in name order:
  - 90/91 Entei & Raikou: `[244, 243]`
  - 92/93 Raikou & Suicune: `[243, 245]`
  - 94/95 Suicune & Entei: `[245, 244]`
- Branch is based on `upstream/master`. Diff is only these six card files. No generated JSON.

## Validation
- `bun run dex:lint` — all Pokémon cards have valid dexId
- set-only scope: `data/HeartGold & SoulSilver/Unleashed/{90-95}.ts`

PR #2364, opened 2026-09-16
fix(compiler): tolerate non-JSON datas.json fetch

lamplis:
## Summary
- Compile was aborting when `https://assets.tcgdex.net/datas.json` returned a CDN error body such as `no available server` (`SyntaxError: Unexpected identifier "no"` in `fetchRemoteFile`).
- `fetchRemoteFile` now checks `resp.ok`, parses JSON only on success, retries 3 times, and does not cache a failed fetch.
- The compile prefetch soft-fails so card/set JSON still compiles if the picture index stays down (picture lookups already catch and skip).
- Bruno GraphQL `cards` queries are scoped to one card so CI does not request the full unbounded list.

Split out of #2327 so that PR stays data-only.

## Test plan
- [ ] `bun run compile` gets past `1. Loading remote sources` even if datas.json is 503/plain text
- [ ] CI: Ubuntu / macOS / Windows `Install deps` (compile) plus Bruno suite


thomas-bassett:
Skipping the fetch on a failure would end up with cards not having an image url attached in the compiled JSON, downstream meaning no url in the API. Hence why this is called at the start and cached. 

**we shouldn't continue compiling if that fails.** 

the retry bit is fine to help cover small blips.

thomas-bassett (changes requested):
as above

PR #2370, opened 2026-09-18
(data) Complete the 30th Celebration sets: variants, marketplace ids, dex, regulation marks, RGB Mews

KlausDerKleber:
Follows up #2361, where the variants were asked for. While adding them a few other gaps in the import turned up, so this closes all of them together.

## Variants and marketplace ids

Neither set carried a `variants` block, so no card referenced a Cardmarket or TCGplayer product.

- **30th Celebration** — all 158 cards, `holo` or `normal` by rarity
- **30th Classic Collection** — all 30 cards, `holo`

`thirdParty.cardmarket` comes from Cardmarket expansion 6601, `thirdParty.tcgplayer` from tcgcsv groups 24722 and 24837. Four cards (143, 151, 152, 154) carry only a Cardmarket id, because TCGplayer has not listed them yet.

**The ids are matched by name, not counted.** Cardmarket's product order for this expansion is not the collection order: Ninetales (010) sits between 128 and 129, one id is skipped outright (907626), and four cards were added days after the rest. Counting positions would have mis-assigned every id from card 10 onward. Each of the 188 products resolves to exactly one card, and no id is used twice. The Classic Collection is numbered in an order of its own here, so those match by card name as well; the two halves of Darkrai & Cresselia LEGEND are told apart by which one carries the attacks — 020 has Lost Crisis, so it is the bottom half.

## The other gaps

From the Bulbapedia card pages and the English set list:

| | |
|---|---|
| `dexId` | added on 155 cards — the three Trainers have none, as they should |
| `regulationMark` | added on all 158 |
| `description` | added on 131 — the ex cards print no Pokédex entry |

**The three RGB Rare Mew prints were missing.** `R`, `G` and `B` reprint 065, so they are derived from that card and keep every language it has. No marketplace lists them yet, so they carry a variant without ids. `RGB Rare` is added to the rarity union next to `Pikachu Rare` and `Futuristic Rare`.

**`30th Classic Collection.ts` had `cardCount.official: 0`.** It is 30.

Verified with `bun run validate` and a full `bun run compile`: 30th now reports 161 cards of 128 official, 30th-c reports 30 of 30.


KlausDerKleber:
@thomas-bassett There you go. Would be great if we can get the 2 new Japanese sets in as well, as for me, personally, these are more important 😬

thomas-bassett (approved):

PR #2374, opened 2026-09-21
data: set Classic Collection rarity on 30th Classic Collection cards

lamplis:
## Changes

Set `rarity: "Classic Collection"` on all 30 cards in **30th Classic Collection** (`30th-c`).

## Details

Follow-up to #2370, which added variants and marketplace ids but left every card at `rarity: "None"`.

Celebrations Classic Collection already uses `Classic Collection` for the same print type (star + C symbol). This aligns `30th-c` with that convention and with the existing rarity union.

Verified with `bun run validate` and `cd server && bun run compile`.


lamplis:
Official rarity symbol attached

Rarity_Classic_Collection
Rarity_Classic_Collection

thomas-bassett:
Could you supply an example card image with that Rarity? for the classic collection the ones i just spot checked all had their original rarity symbols on them. Which means the rarity is still wrong anyway I just want to figure out what they should be properly

lamplis:
> Could you supply an example card image with that Rarity? for the classic collection the ones i just spot checked all had their original rarity symbols on them. Which means the rarity is still wrong anyway I just want to figure out what they should be properly

Here is the official list from pokemon  [160_30th_anniversary_celebrations_set_checklist.pdf](https://cdn.shopify.com/s/files/1/0663/9612/7517/files/160_30th_anniversary_celebrations_set_checklist.pdf)

It won't be visible in the reprinted cards but it's the official pokemon rarity name and symbol.

thomas-bassett:
> > Could you supply an example card image with that Rarity? for the classic collection the ones i just spot checked all had their original rarity symbols on them. Which means the rarity is still wrong anyway I just want to figure out what they should be properly
> 
> Here is the official list from pokemon [160_30th_anniversary_celebrations_set_checklist.pdf](https://cdn.shopify.com/s/files/1/0663/9612/7517/files/160_30th_anniversary_celebrations_set_checklist.pdf)
> 
> It won't be visible in the reprinted cards but it's the official pokemon rarity name and symbol.

Thank you, I see what you mean now. 

This is a oddly specific use case as the data on the cards ideally should be the source of truth. i.e the cards have the old rarity. 

Im personally of the opinion that we should stick with the cards, even though they technically don't match for "pull" rarity. if you disagree i think we should bring this into the discord for a discussion and see what the majority thinks. Will leave it open for you to decide

lamplis:
Thx @thomas-bassett for the feedback.

For collecting, following official pokemon set list rarities make things simple in term of enduser user experience. 
Otherwise you should consider theses cards to be cel30 stamped variants of their original cards buried in old collections like we do with specials cards reprinted out of a set. This would be even more accurate but again, bad user experience.
If we keep the original rarity, why not keeping the original parent set for the same reason.

We have to decide to go simple like pokemon is doing or do strict card variant management.

I honestly prefer a simple list to display master sets in one shot without requesting all variants with a specific stamp for this specific set only and prioritize consistency (6 to 8 rarities per set and variants not part of sets like pop cards.)

Personnaly, I think we should have only one 30th anniversary set and have the 30 cards with Classic Collection rarity reflecting official pokemon offer.

But I'll follow the vote.

thomas-bassett:
> Thx @thomas-bassett for the feedback.
> 
> For collecting, following official pokemon set list rarities make things simple in term of enduser user experience. Otherwise you should consider theses cards to be cel30 stamped variants of their original cards buried in old collections like we do with specials cards reprinted out of a set. This would be even more accurate but again, bad user experience. If we keep the original rarity, why not keeping the original parent set for the same reason.
> 
> We have to decide to go simple like pokemon is doing or do strict card variant management.
> 
> I honestly prefer a simple list to display master sets in one shot without requesting all variants with a specific stamp for this specific set only and prioritize consistency (6 to 8 rarities per set and variants not part of sets like pop cards.)
> 
> Personnaly, I think we should have only one 30th anniversary set and have the 30 cards with Classic Collection rarity reflecting official pokemon offer.
> 
> But I'll follow the vote.

yeah i get what your saying, some of the cards in the classic collection are not a 1/1 reprint though. i believe the charizard has a different border. this dosnt counter your point though because it could still be a variant

theres certainly an argument to be made both ways, i think we should go for a hybrid approach in the long term. i.e. if a card is released as part of a set it should stay as a variant of any of the same card. But if its a reprint like these we have a "linked" on both card objects that denotes its a reprint. That would give us the best of both worlds then

another example would be the trick and trade sets, we dont have those as sets yet and they are just put as variants in exsisting sets

